PR #101, opened 2017-07-11
Update missing UIA mappings

boggydigital:
This change should cover all existing gaps in the UIA mapping for states, roles, attributes, etc.

Most of the information in the mappings come from existing Microsoft Edge implementation, verified by Inspect.exe tool (part of Windows SDK). For other changes I've tried to map as close as possible to preexisting IA2/MSAA/ATK/AX mapping whenever it makes sense.

Finally there are just a couple mappings defined as "Not mapped" - this reflects the fact that it is not mapped today, nor there is an existing UIA concept to map it to.

### DIFF
--- a/index.html
+++ b/index.html
@@ -805,8 +805,10 @@
                 </td>
                 <td class="uia">
                     <div class="general">
-                        ?TODO: look in more detail at UIA table implementation
-                    </div>
+                      <div class="ctrltype">
+                        <span class="type">Control Type: </span><code>Group</code>
+                      </div>
+                      <div class="ctrltype"><span class="type">Localized Control Type:</span> <code>"colgroup"</code></div>                    </div>
                 </td>
                 <td class="atk"><div class="general">Not mapped</div></td>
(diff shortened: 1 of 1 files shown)


jasonkiss:
Thanks, @boggydigital. 

As figuring out from the diff exactly which elements and atts have had their mappings update, I'm noting here that the following were updated in this PR:

## Elements
- colgroup
- details
- embed
- pre
- q
- rp
- rt
- ruby

## Attributes
- abbr
- contenteditable
- controls
- coords
- datetime
- draggable
- dropzone
- headers
- height
- hreflang
- low
- max
- maxlength
- min
- multiple for input
- pattern
- readonly
- rows
- scope
- size
- span
- spellcheck
- src
- start
- tabindex
- title on html elements
- title on abbr, dfn
- title on menuitem
- type on button
- type on embed, object, script, source, style
- type on menu
- type on menuitem
- type on ol
- usemap
- value for meter, progress
- width

PR #103, opened 2017-08-17
Move ValuePattern for a[href] from Core-AAM

boggydigital:
Please see https://github.com/w3c/aria/pull/633

ARIA doesn't provide href attribute, so it makes sense to add it here

### DIFF
--- a/index.html
+++ b/index.html
@@ -222,7 +222,13 @@
                 <th><a href="https://www.w3.org/TR/html/textlevel-semantics.html#the-a-element"><code>a</code></a> <span class="el-context">(represents a <a href="https://www.w3.org/TR/html/links.html#hyperlink">hyperlink</a>)</span></th>
                 <td class="aria"><a class="core-mapping" href="#role-map-link"><code>link</code></a> role </td>
                 <td class="ia2"><div class="general">Use WAI-ARIA mapping</div></td>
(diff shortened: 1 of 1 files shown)


joanmarie:
FWIW: The issue being addressed in this PR isn't limited to UIA. ATK and IA2 also expose the URI for "real" links (e.g. the `a` element with an `href` attribute), and cannot do so for ARIA links until we have role and property parity. :slightly_frowning_face: Thus if the conclusion of the Editors is that the HTML AAM shouldn't just point to the ARIA mappings for the `a` element, I think it makes sense to make that change across the board.

That said, the HTML AAM seems to already have [mappings for exposure of the `href` attribute.](https://w3c.github.io/html-aam/#att-href). Those mappings just seem to have gotten a tad on the broken side. But once that's fixed (see #105), you'll have the mapping you need for the href attribute (Currently verifiable here: https://joanmarie.github.io/html-aam/#att-href).

Long way of saying the following: One approach is this pull request. Another approach would be to leave the HTML AAM as-is (i.e. pointing to the ARIA mappings) and just remove the mention of the `href` attribute exposure from the Core AAM.

boggydigital:
Joanmarie is right and I completely forgot about the attribute mapping - so this is absolutely not needed.

PR #105, opened 2017-08-17
Remove left-over UIA Express mappings from State and Property table

joanmarie:
The HTML-AAM no longer includes mappings for UIA Express. In addition,
the presence of these left-over mappings in the State and Property
Mappings table results in broken mappings when viewing the content by
state/property.


joanmarie:
I've just linked my github account to address the failing check.

To see an example of the brokenness being addressed, look at https://w3c.github.io/html-aam/#att-href, in which MSAA + IA2 and ATK state "The value of the href attribute is stored in the Value.Value UIA property." And the UIA mapping -- the one place the aforementioned statement belongs -- lacks it. Plus there's an "undefined" row after the "Comments" row.

stevefaulkner (approved):

PR #107, opened 2017-08-29
Issue 106: The <form> element should only map to FORM landmark with restriction of accessible name 

jongund:
Updated the form element mapping to only be a form landmark when it html as an accessible name

### DIFF
--- a/index.html
+++ b/index.html
@@ -1387,7 +1387,7 @@
             </tr>
             <tr tabindex="-1" id="el-form">
                 <th><a href="https://www.w3.org/TR/html/sec-forms.html#the-form-element"><code>form</code></a></th>
-                <td class="aria"><a class="core-mapping" href="#role-map-form "><code>form </code></a>role</td>
+                <td class="aria"><a class="core-mapping" href="#role-map-form "><code>form</code></a> role if the <a href="https://www.w3.org/TR/html5/forms.html#the-form-element"><code>form</code></a> element has an <a class="termref">accessible name</a>. Otherwise, no corresponding role.</td>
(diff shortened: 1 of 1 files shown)


jongund:
I am working on some landmark rules with Deque labs and having this requirement in the spec would similify rules relating to top level landmarks (e.g. main, banner and contentinfo)

stevefaulkner (approved):

stevefaulkner:
thanks @jongund

joanmarie:
What does "no corresponding role" mean with respect to platforms? Does it mean "pruned from the tree" or is it mapped to a platform accessibility role? I don't have strong feelings either way, as long as it doesn't show up as ATK_ROLE_UNKNOWN on my platform. :smile: But I do think more guidance is called for in order to ensure implementors know what to do for nameless HTML `form` elements.

Speaking just for my platform, if it is not pruned from the accessibility tree, possible mappings would be:

* `ATK_ROLE_SECTION` (which is the mapping used for the `div` element and other generic block elements which need to remain in the accessibility tree but have no semantic meaning)
* `ATK_ROLE_FORM` (which is not the same as the [mapping for the ARIA `form` role](https://rawgit.com/w3c/aria/master/core-aam/core-aam.html#role-map-form), which should be exposed as an `ATK_ROLE_LANDMARK`).

Thoughts?

EDIT: Compare to what was done for the `footer` element which has "no corresponding role" in the ARIA row, AND explicit platform mappings beneath.

stevefaulkner:
@joanmarie  I will add extra suggested mappings and we can take it from there, thanks!

stevefaulkner:
fixed via 2724b297db1f0bc129bc4fdf711e6c48cf9ac1e4

PR #108, opened 2017-10-10
Update index.html

stevefaulkner:
updated `<form>` mapping to reflect how `<section>` is defined with and without accessible name
/cc @joanmarie @asurkov @cookiecrook  @boggydigital

### DIFF
--- a/index.html
+++ b/index.html
@@ -1394,6 +1394,53 @@
                 <td class="ax"><div class="general">Use WAI-ARIA mapping</div></td>
                 <td class="comments"></td>
             </tr>
+             <tr tabindex="-1" id="el-form">
+                  <th><a href="https://www.w3.org/TR/html/sec-forms.html#the-form-element"><code>form</code></a></th>
+                  <td class="aria">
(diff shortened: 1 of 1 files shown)


joanmarie:
If the general consensus is that a `form` without an accessible name might as well be a `div`, then what you've done for ATK makes perfect sense.

On the other hand, if the general consensus is that a `form` without an accessible name is still something ATs should be able to distinguish from a `div`, then I think the ATK mapping would be to ATK_ROLE_FORM.

(At the moment, I don't have much an opinion about the consensus question; just providing the mappings for each scenario. :grin: )

jasonkiss:
I like the idea of using ATK_ROLE_FORM for a `form` with no accessible name, and ATK_ROLE_LANDMARK where the `form` does have an accessible name. After all, the `form` element exists for a reason, it won't always deserve to be a landmark, and both IA2 and ATK have explicit form roles.

IA2 also has IA2_ROLE_LANDMARK, but I note that CORE-AAM maps the form role to IA2_ROLE_FORM, and not IA2_ROLE_LANDMARK. I don't expect consistency across these or any a11y APIs, but it would be nice, especially with these two where it makes sense. 

In CORE-AAM, UIA maps the form role more or less as you've expressly added it in your PR,  @stevefaulkner, except for the Landmark Type ("Form" in CORE, but "Custom" in this PR). I would've thought we could just point to the WAI-ARIA mapping for UIA as well here?

stevefaulkner:
fixed via 2724b297db1f0bc129bc4fdf711e6c48cf9ac1e4

Issue #24, opened 2016-10-12
HTML-AAM: consider adding note to style element that some CSS will change accessibiltiy tree


jasonkiss:
_From @cyns on January 12, 2016 23:50_

For example, display:none removes elements from the accessibility tree.

_Copied from original issue: w3c/aria#198_

asurkov:
Do we want to handle CSS styles mapping as part of HTML-AAM at all? If so, then we should consider to add a new section 'CSS styles mapping' like we have 'HTML elements mapping' and 'HTML attributes mapping'. Should we have a separate issue for this? @jasonkiss

jasonkiss:
I believe there may still be an intention for something like a CSS-AAM, which would be my preference. A note regarding impact of some CSS in the `style` element makes sense, but I don't think we should start adding mappings for other technologies.

jasonkiss:
Fixed with PR #114, thanks to @scottaohara.

PR #114, opened 2018-01-21
add note for style element

scottaohara:
[per issue 24](https://github.com/w3c/html-aam/issues/24), made a draft at a note to mention how CSS can affect the accessibility tree.

### DIFF
--- a/index.html
+++ b/index.html
@@ -3045,10 +3045,7 @@
                 <td class="ax"><div class="general">Not mapped</div></td>
                 <td class="comments">
                   <div class="general">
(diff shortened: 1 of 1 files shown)


jasonkiss:
Thanks for the PR, @scottaohara. Would you be willing to rewrite as follows and resubmit?

**Note:** There are instances where CSS properties can affect what is exposed by accessibility APIs. For instance, `display:none` or `visibility:hidden` will remove an element from the accessibility tree and hide its presence from assistive technologies.

scottaohara:
thanks for the feedback!  updated with the new wording.

jasonkiss:
Awesome. Thanks!

PR #127, opened 2018-03-09
Add UIA mapping for the placeholder attribute

boggydigital:
Per MSDN - the placeholder is best mapped to HelpText property: https://msdn.microsoft.com/en-us/library/dd757479(v=vs.85).aspx


jasonkiss:
Thanks, @boggydigital. Will the CORE mapping for aria-placeholder also be getting updated to HelpText for UIA, in which case we could leave the HTML mapping as is?

boggydigital:
Pardon @jasonkiss, missed this somehow. You're correct and I've opened https://github.com/w3c/core-aam/pull/6 to do just that. I'll close that PR now

PR #146, opened 2018-08-30
Clarify which legend/figcaption/caption element to use

### LINKED ISSUE #145: Unclear which legend element to be used as accessible name for fieldset
> 5.5.1 fieldset Element Accessible Name Computation
> If the fieldset element has an aria-label or an aria-labelledby attribute the accessible name is to be calculated using the algorithm defined in Accessible Name and Description: Computation and API Mappings 1.1.
> Otherwise use legend subtree
> Otherwise use title attribute
> If none of the above yield a usable text string there is no accessible name

What is the legend subtree?

Similarly for figure/figcaption, table/caption.

Cases to consider:

* there might be no such child element
* there might be more than one such child element
* the [flat tree](https://drafts.csswg.org/css-scoping/#flat-tree) might differ from the DOM tree (e.g. if there's a `display: contents` element in between, or if it crosses a shadow DOM boundary).
* the element might exist in the DOM but not in the flat tree (e.g. `display: none`).

zcorpan:
Fixes #145.

### DIFF
--- a/index.html
+++ b/index.html
@@ -5583,8 +5583,8 @@
         <h4><code>fieldset</code> Element Accessible Name Computation</h4>
         <ol>
           <li>If the <code>fieldset</code> element has an <a href="https://www.w3.org/TR/wai-aria-1.1/#aria-label"><code>aria-label</code></a> or an <a href="https://www.w3.org/TR/wai-aria-1.1/#aria-labelledby"><code>aria-labelledby</code></a> attribute the <a class="termref">accessible name</a> is to be calculated using the algorithm defined in <a href="" class="accname">Accessible Name and Description: Computation and API Mappings 1.1</a>. </li>
-          <li>Otherwise, if the <code>fieldset</code> element has a <a href="https://dom.spec.whatwg.org/#concept-tree-child">child</a> that is a <code>legend</code> element, then use the subtree of the first such element</li>
-          <li>Otherwise, if the <code>fieldset</code> element has a <code>title</code> attribute, then use that attribute</li>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -5583,9 +5583,9 @@ <h3><code>fieldset</code> and <code>legend</code> Elements</h3>
         <h4><code>fieldset</code> Element Accessible Name Computation</h4>
         <ol>
           <li>If the <code>fieldset</code> element has an <a href="https://www.w3.org/TR/wai-aria-1.1/#aria-label"><code>aria-label</code></a> or an <a href="https://www.w3.org/TR/wai-aria-1.1/#aria-labelledby"><code>aria-labelledby</code></a> attribute the <a class="termref">accessible name</a> is to be calculated using the algorithm defined in <a href="" class="accname">Accessible Name and Description: Computation and API Mappings 1.1</a>. </li>
```
mcking65, line 5585: I actually do not understand why there is any remaining need for conditional language here. Might this be a hold-over from back in the days when the accessible name computation spec was an "ARIA-only" thing buried inside the UAIG?

In other words, why are there naming algorithms being defined outside the accname spec? The accname spec defines when to use content or title. So, why can't the accessible name calculation be used all the time? If every element has its own algorithm, testing name computation will continue to be unnecessarily complex, or perhaps more accurately, impossibly complex.

@accdc, @joanmarie, @jnurthen, does this make sense to you?

zcorpan: I can envision there being a generic algorithm that applies to all HTML/SVG elements, which has an element-specific hook that `fieldset`, `table` etc can define to let `legend`/`caption` go in between ARIA attributes and the `title` attribute. But this is basically an editorial detail.

stevefaulkner (approved):

PR #153, opened 2018-10-22
fix: Ran Nu Validator to fix minor issues

nschonni:
Because GitHub is having issues, I did this with by looking at https://validator.w3.org/nu/?doc=https%3A%2F%2Fw3c.github.io%2Fhtml-aam%2F#l3392c17
I can split up the commits if that is better


nschonni:
Marked as non substantive for IPR from ash-nazg.

scottaohara:
Looks good, thank you!

PR #187, opened 2019-05-21
176 map and area updates

### LINKED ISSUE #176: There's no name computation for the area element
`area` isn't listed in the Accessible Name and Description Computation
https://www.w3.org/TR/html-aam-1.0/#accessible-name-and-description-computation

although it is listed in HTML Element Role Mappings
https://www.w3.org/TR/html-aam-1.0/#el-area

```
<img src="usa.png" alt="Map of United States" usemap="#us_map">
<map name="us_map">
 <area alt="Alabama" href="alabama.htm" shape="poly" coords="..." />
 <area alt="Alaska" href="alaska.htm" shape="poly" coords="..." />
 ....
 <area alt="Wyoming" href="wyoming.htm" shape="poly" coords="..." />
</map>
```

There are two different ALT's involved here - one on the IMG referencing the MAP AREAs and the ALT on each AREA -  and there could be null ALTs on either element (or both elements). This poses a few questions:

1) Should the accessible name of an AREA be a combination of both IMG ALT and AREA ALT? That could get very verbose with the example above when reading each AREA link.
2) If not a combination then should IMG ALT be exposed in some other way, such as mapping IMG ALT to AREA link accessible description?
3) What should happen when IMG has null ALT, but IMG USEMAP references MAP AREAs with non-null ALT? Should it set the image to role=presentational, even though the USEMAP attribute says the image isn't presentational? If AREA is voiced as a combination of IMG ALT and AREA ALT then there will be cases where you don't want any IMG ALT to reduce verbosity.
4) What should happen when IMG has non-null ALT, but AREA has a null ALT?

To get an idea of UA differences currently:

- NVDA 2018 with FF60 voices IMG ALT and AREA ALT as link title
- JAWS 2018 with IE11 voices AREA ALT as link title
- JAWS 2018 with FF60 voices AREA HREF as link title
- VoiceOver with Safari on macOS 10.13 voices AREA ALT as link title
- VoiceOver with Safari on iOS 11.4 voices AREA ALT as link title, but only reads IMG ALT (and not AREA ALT) when the speak screen gesture is used (i.e. no link roles read, and no indication AREA links exist - the only way to find the AREA links seems be setting rotor is set to Links, and using the move to next rotor item gesture )

MDN page here:
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/area

WCAG technique here:
https://www.w3.org/TR/WCAG20-TECHS/H24.html

Tested support here:
https://www.powermapper.com/tests/screen-readers/labelling/area-alt/
https://www.powermapper.com/tests/screen-readers/labelling/area-alt-img-null-alt/
https://www.powermapper.com/tests/screen-readers/labelling/area-title/
https://www.powermapper.com/tests/screen-readers/labelling/area-aria-label/
https://www.powermapper.com/tests/screen-readers/labelling/area-aria-labelledby/

scottaohara:
Closes #176

Adds `area` accessible name and accessible description computation steps.

Updates `map`'s mapping to match what is reported by AXAPI in different situations.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/html-aam/pull/187.html" title="Last updated on May 21, 2019, 3:15 PM UTC (fac03a7)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/html-aam/187/87236e8...fac03a7.html" title="Last updated on May 21, 2019, 3:15 PM UTC (fac03a7)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -5866,7 +5866,7 @@
         <h4><code>img</code> Element Accessible Name Computation</h4>
         <ol>
           <li>
-            If the <code>img</code> element  has an <a href="https://www.w3.org/TR/wai-aria-1.1/#aria-label"><code>aria-label</code></a> or an <a href="https://www.w3.org/TR/wai-aria-1.1/#aria-labelledby"><code>aria-labelledby</code></a> attribute the <a class="termref">accessible name</a> is to be calculated using the algorithm defined in <a href="" class="accname">Accessible Name and Description: Computation and API Mappings 1.1</a>.
+            If the <code>img</code> element has an <a href="https://www.w3.org/TR/wai-aria-1.1/#aria-label"><code>aria-label</code></a> or an <a href="https://www.w3.org/TR/wai-aria-1.1/#aria-labelledby"><code>aria-labelledby</code></a> attribute the <a class="termref">accessible name</a> is to be calculated using the algorithm defined in <a href="" class="accname">Accessible Name and Description: Computation and API Mappings 1.1</a>.
           </li>
           <li>Otherwise use <code>alt</code> attribute.</li>
           <li>Otherwise use <code>title</code> attribute.</li>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1,17 +1,17 @@
...
 <head>
-  <meta charset="utf-8" />
+  <meta charset="utf-8">
   <title>HTML Accessibility API Mappings 1.0</title>
   <!-- Temporary replacement of these 2 links to CSS resources at rawgit.com with local sources until common solution for resources like these shared among different AAMs is found (note two arrow images added to local img/ dir referenced from mapping-tables.css)-->
   <!--
-  <link href="https://rawgit.com/w3c/aria/master/common/css/mapping-tables.css" rel="stylesheet" />
-  <link href="https://rawgit.com/w3c/aria/master/common/css/common.css" rel="stylesheet" />
+  <link href="https://rawgit.com/w3c/aria/master/common/css/mapping-tables.css" rel="stylesheet">
+  <link href="https://rawgit.com/w3c/aria/master/common/css/common.css" rel="stylesheet">
```
stevefaulkner, line 9: any idea why we are pointing to "https://rawgit.com/..."

stevefaulkner: scratch that saw note below
  `<!-- Temporary replacement of these 4 script references to JS resources at rawgit.com with local sources until common solution for resources like these shared among different AAMs is found -->`

stevefaulkner (approved):

index.html
```diff
@@ -1,17 +1,17 @@
 <!DOCTYPE html>
 <html lang="en-US" xmlns="http://www.w3.org/1999/xhtml">
```
marcoscaceres: While we are here... xmlns ✂️

index.html
marcoscaceres: Is this not on a gh-pages branch?

index.html
stevefaulkner: killjoy

marcoscaceres: Heh, we should easter-egg it... make it xhtml2

marcoscaceres:
We should set up some time to chat about how to use `data-cite` and automatic cross references in ReSpec. We should never need to include absolute URLs in the spec, specially ones to the HTML spec. 

Happy to do some training for folks here!

scottaohara:
I think that's a good idea @marcoscaceres

PR #193, opened 2019-05-23
Editorial: cross ref cleanup

marcoscaceres:

 * [X] Confirmed there are no ReSpec/BikeShed errors or warnings.

Cleans up cross references and starts citing HTML. 

Dropped keygen as it's no longer a thing.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/html-aam/pull/193.html" title="Last updated on May 23, 2019, 12:55 PM UTC (63ce25c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/html-aam/193/51f0f55...63ce25c.html" title="Last updated on May 23, 2019, 12:55 PM UTC (63ce25c)">Diff</a>

### DIFF
--- a/biblio.js
+++ b/biblio.js
@@ -1,40 +1,4 @@
 var biblio = {
-	
-	"ACCNAME-AAM": {
-		"title": "Accessible Name and Description: Computation and API Mappings 1.1",
-		"href": "http://www.w3.org/TR/accname-aam-1.1/",
-		"status": "WD",
-		"publisher": "W3C",
-		"authors":  [
-			"Joseph Scheuhammer",
-			"Michael Cooper",
-			"Andi Snow-Weaver",
-			"Aaron Leventhal"
-		],
-		"etAl": true,
-		"deliveredBy":  [
-			"http://www.w3.org/WAI/ARIA/"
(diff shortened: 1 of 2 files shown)


scottaohara (commented):
@marcoscaceres I went through the HTML elements and did some additional cleanup so that links weren't being populated into the `summary` element.

There are still a bunch of link refs that need to be cleaned up in the attribute tables.  I [pushed a quick attr update for 2 instances](https://github.com/w3c/html-aam/pull/193/commits/63ce25c8efa479bfd5ff9f1fccb984ceda8ddd8c) to this branch, could you please verify that is how the other attributes refs should be fixed?

marcoscaceres:
> could you please verify that is how the other attributes refs should be fixed?

They look good, but let’s take it slow. We can do this over a couple of PRs.

marcoscaceres:
Thanks for your help @scottaohara. I'm going to set up some validation (which will initially fail spectacularly) so we have a good idea what we need to do. With a bit of work, we should be able to get the spec to a good state and prevent it from getting stale cross references going forward 🤞

PR #206, opened 2019-06-06
Update UIA mappings for ins and del

melanierichards:
What it says on the tin :)

For progress on #141

The following tasks have been completed:

 * [x] Confirmed there are no ReSpec/BikeShed errors or warnings.
 * [ ] Modified Web platform tests (link to pull request)

Implementation commitment:

 * [ ] WebKit (https://bugs.webkit.org/show_bug.cgi?id=)
 * [x] Chromium (https://bugs.chromium.org/p/chromium/issues/detail?id=)
 * [ ] Gecko (https://bugzilla.mozilla.org/show_bug.cgi?id=)

***
<a href="https://pr-preview.s3.amazonaws.com/melanierichards/html-aam/pull/206.html" title="Last updated on Jun 6, 2019, 9:43 PM UTC (90a7395)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/html-aam/206/b71b317...melanierichards:90a7395.html" title="Last updated on Jun 6, 2019, 9:43 PM UTC (90a7395)">Diff</a>

PR #213, opened 2019-06-13
Update mappings for ins and del elements: UIA, IA2, ATK

### LINKED ISSUE #141: Add mappings for ins and del
The `ins` and `del` elements currently are shown as having no accessible object for three of the four platforms. In addition, for ATK it states that exposure should be done via text attribute.

In order to easily support presentation (or exclusion, should the user prefer) of edits marked up via `ins` and `del`, I think there should be an accessible object. IAccessible2 has just added roles for each of these: https://github.com/LinuxA11y/IAccessible2/blob/master/api/AccessibleRole.idl#L307. ATK and AT-SPI2 will soon, hopefully: https://gitlab.gnome.org/GNOME/atk/issues/2.

In terms of the mappings for ATK, until the new roles have been created, I think using ROLE_SECTION makes sense (i.e. the same mapping as `div` because `ins` and `del` can have child elements.) 

We still need a way to make it easy for ATs in Linux to find these elements. A text attribute (be it formatting or something like `diff:del`/`diff:ins`) doesn't make sense to me as Orca doesn't bother getting text attributes unless a user explicitly asks for them via command. Checking all text attributes on the slim chance there's an `ins` or `del` will not be performant. Orca does, for a variety of reasons, check object attributes. Thus I think exposure via object attribute is the way to go. Could we use `xml-roles` for that? If not, an object attribute like `diff:del` and `diff:ins` should work.

scottaohara:
Closes #141

Updates `del` and `ins` mappings for:
* [UIA implementation commitment](https://github.com/w3c/html-aam/pull/206)
* [IA2](https://github.com/LinuxA11y/IAccessible2/blob/master/api/AccessibleRole.idl#L307)
* [ATK open issue to implement](https://gitlab.gnome.org/GNOME/atk/issues/2)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/html-aam/pull/213.html" title="Last updated on Jun 13, 2019, 12:34 PM UTC (dd70111)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/html-aam/213/153f948...dd70111.html" title="Last updated on Jun 13, 2019, 12:34 PM UTC (dd70111)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -701,7 +701,9 @@
                 <td class="comments"></td>
             </tr>
             <tr tabindex="-1" id="el-canvas">
-                <th><a data-cite="HTML">`canvas`</a></th>
+                <th>
+                  <a data-cite="HTML">`canvas`</a>
+                </th>
                 <td class="aria">No corresponding role</td>
                 <td class="ia2">
                   <div class="role">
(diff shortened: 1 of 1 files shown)


joanmarie:
So regarding ATK/AT-SPI2. I took this issue as a prompt to do the API changes for ATK and AT-SPI2. :smile: I, of course, need code review and approval, but I don't think that will be a problem.

Once the new roles land in the APIs, then we'll need the user agents to use the new ATK roles. With this in mind, I'm not sure if you want to have the mappings point to the new ATK roles -- which will ultimately be implemented -- or point to ATK_ROLE_SECTION as you are currently.

scottaohara:
@joanmarie I expected another update to these mappings once the ARIA equivalent roles land as well, so we can change to more specific ATK / AT-SPI2 roles once they land.   That work?

joanmarie:
> @joanmarie I expected another update to these mappings once the ARIA equivalent roles land as well, so we can change to more specific ATK / AT-SPI2 roles once they land. That work?

Sure. I just wanted to update you so that you know we (GNOME, also known as Joanie) are finally moving forward on this so you're aware your spec will be needing an update. (In case I forget to tell you later.... Premature senility and all that. ;))

stevefaulkner (approved):
LGTM

Issue #189, opened 2019-05-21
Add HTML elements


scottaohara:
The following elements are in the HTML Living Standard but are not represented here.

- [ ] [hgroup](https://html.spec.whatwg.org/multipage/sections.html#the-hgroup-element)
- [ ] [slot](https://html.spec.whatwg.org/multipage/sections.html#the-slot-element)
- [ ] [autonomous custom element](https://html.spec.whatwg.org/multipage/custom-elements.html#autonomous-custom-element)

scottaohara:
This issue will be picked up once #226 is completed.

scottaohara:
copied from ARIA in HTML issue regarding custom elements:
Per: HTML issue 4658 and ARIA issue 984, it should be specified that a custom element may have no corresponding role (map to generic) or may have a role set by `ElementInternals`. 

If the role is set in that way, and the custom element doesn't merely act as a container for child elements (light or shadow dom) with their own roles, then the listing for the custom element would need to indicate that it either maps to generic, OR the role set by `ElementInternals`, and to reference THOSE roles for appropriate mappings.

carmacleod:
Can we create a separate issue for `hgroup` so that it is not held up waiting for autonomous custom elements? Is hgroup "Not mapped" everywhere? Does it need to wait for the "If mapped" PR #226?

(For context, I'm hoping to point to `hgroup` (and `slot`) from the html spec, but at present, they're missing from html-aam). :)

scottaohara:
I do not think #226 should be a blocker here anymore, per my comment today in that issue's WIP PR.

`hgroup` is not mapped anywhere and may well continue to be unmapped (though impact the heading levels of contained elements) depending on how that gets reworked in the HTML spec.

`slot` is not mapped, nor should it be.

autonomous custom elements are either `role=generic` or mapped to the role declared via its `ElementInternals`.

form associated custom elements should map to the form-related role declared via its `ElementInternals`.

PR #226, opened 2019-07-23
if mapped initial commit

### LINKED ISSUE #151: Need language/mapping for "not-mapped" elements which must be exposed due to attributes
Consider the following:
    `<b tabindex="0">foo</b>`
Because it is focusable, it has to have an accessible object (i.e. be mapped) so that accessibility events can be emitted.

Likewise: `<b title="hello world!">foo</b>` would need an accessible object because otherwise there's no way to expose and associate the title value.

At the present time, the HTML-AAM lacks language/mapping for a number of inline elements. There should be either language (a normative note) or mappings for how to expose "not mapped" elements which must be exposed due to attributes.

As for what the mappings should be, a quick glance makes me think that the content for "a (no href attribute)" is the expected result for all platforms for the sorts of things I've described.

scottaohara:
* adds bullet to element role mappings
* adds TODO add if mapped for ia2/uia
* adds if mapped conditional for ATK mappings
* minor cleanups

Work to be done:
- [ ] similar wording and mapping expectations for IA2 / UIA (see todo comments)
- [x] revise wording for `div` & any other 'block' elements that should be updated to match

Closes #151 

 * [x] Confirmed there are no ReSpec/BikeShed errors or warnings.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/html-aam/pull/226.html" title="Last updated on Jul 28, 2019, 3:29 PM UTC (a0ae511)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/html-aam/226/f7c394a...a0ae511.html" title="Last updated on Jul 28, 2019, 3:29 PM UTC (a0ae511)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -536,7 +536,8 @@
                   <span class="type">Role:</span> `ATK_ROLE_STATIC` (if mapped)
                 </div>
                 <div class="general">
-                  Otherwise no accessible object. Exposed as "font-weight" text attribute on the text container. The value depends on the platform.
+                  Otherwise no accessible object.
+                  Exposed as "font-weight" text attribute on the text container. The value depends on the platform.
                 </div>
               </td>
               <td class="ax">
@@ -584,7 +585,8 @@
                   <span class="type">Role:</span> `ATK_ROLE_STATIC` (if mapped)
                 </div>
                 <div class="general">
-                  Otherwise no accessible object. May affect on "writing-mode" text attribute on its text container.
+                  Otherwise no accessible object.
+                  May affect on "writing-mode" text attribute on its text container.
                 </div>
               </td>
               <td class="ax"></td>
@@ -613,7 +615,8 @@
                   <span class="type">Role:</span> `ATK_ROLE_STATIC` (if mapped)
                 </div>
                 <div class="general">
-                  Otherwise no accessible object. Exposed as "writing-mode" text attribute on its text container.
+                  Otherwise no accessible object.
+                  Exposed as "writing-mode" text attribute on its text container.
                 </div>
               </td>
               <td class="ax">
@@ -749,12 +752,12 @@
                   </div>
                 </td>
                 <td class="uia">
-                    <div class="ctrltype">
-                        <span class="type">Control Type: </span><code>Image</code>
-                    </div>
-                    <div class="general">
-                        Descendants of the <code>canvas</code> element are mapped separately.
-                    </div>
+                  <div class="ctrltype">
+                    <span class="type">Control Type: </span><code>Image</code>
+                  </div>
+                  <div class="general">
+                    Descendants of the <code>canvas</code> element are mapped separately.
+                  </div>
                 </td>
                 <td class="atk">
                   <div class="role">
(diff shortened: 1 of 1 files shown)


scottaohara:
Initial draft, please take a look if you can @joanmarie @carmacleod @stevefaulkner to make sure i'm on the right track with this.

More work needs to happen here, but I wanted to at least get the verbage in for ATK from @joanmarie.  Needs to add ia2/uia for other span-like elements and need to update the wording for elements like `div` that are "not mapped" / "if mapped"

carmacleod:
Hi, @scottaohara !
There's something funny with the Preview and Diff links - they don't include the table(s)... and it's a little bit rough to review the changes in the code because one needs to expand previous lines of code to figure out which element's table row each change is in... :)
Is it easy to fix the missing table in the Preview | Diff links?
Or do you have any other preview/diff links?

That said, I think it's on the right track from what I could see.

scottaohara:
@marcoscaceres might the preview/diff be an issue with the custom script that converts the giant table into individual element disclosure widgets?

just wonder if there might be a way to fix that / at the very least look into just exposing the full table in this view.

stevefaulkner (approved):

index.html
```diff
@@ -3184,10 +3276,10 @@ <h3>HTML Element Role Mappings</h3>
                 <th><a data-cite="HTML">`small`</a></th>
                 <td class="aria">No corresponding role</td>
                 <td class="ia2">
-                  <!-- add iff mapped -->
                   <div class="general">No accessible object.</div>
+                  <!-- add if mapped -->
```
carmacleod, line 3280: might want to add TODO so you find this later :)

index.html
```diff
@@ -3535,23 +3724,55 @@ <h3>HTML Element Role Mappings</h3>
...
-                    "text-underline-style:solid" text attribute on its text container.
-                  </div>
-                </td>
-                <td class="ax">?</td>
-                <td class="comments"></td>
+              <th><a data-cite="HTML">`u`</a></th>
+              <td class="aria">No corresponding role</td>
+              <td class="ia2">
+                <div class="general">
+                  If mapped:
```
carmacleod, line 3731: I don't think the MSAA or UIA columns for `u` need the "If mapped: ... Otherwise".
I think they should be similar wording/formatting to the MSAA and UIA columns for `s`?

carmacleod:
Can I help with this? I can open a new issue for my last review comment (about u element) if that's holding things up.

scottaohara:
Hi @carmacleod 

No your comment wasn't holding things up.  That was just a part of the spec I hadn't finished updating yet.

It's more that as time has passed, I've become unsure that this is the direction this should be taking. Especially with how things have progressed with ARIA, and many of these elements should have a mapping of `role=generic`.

carmacleod:
Ah, of course. Thanks, @scottaohara 
I was asking because the issue regarding adding hgroup and slot to HTML AAM says https://github.com/w3c/html-aam/issues/189#issuecomment-517949442.

scottaohara:
closing (but not deleting) this pull request as mappings to `generic` will make this unnecessary.

PR #239, opened 2019-09-07
updated guidance regarding whitespace and element naming

### LINKED ISSUE #238: Whitespace alone should be ignored in accName algorithms
We need to make it clear in the specification that attributes that contribute to the accessible name of elements should be ignored if, when white space is trimmed, they return the empty string.

e.g. `alt=" "` should be treated the same as `alt=""`.

scottaohara:
Closes #238 

- [x] Confirmed there are no ReSpec/BikeShed errors or warnings.

Test cases showing current browser support/behavior by some screen readers:
https://scottaohara.github.io/tests/html-img/whitespace.html

Implementation commitment:

 * [ ] WebKit (...)
 * [ ] Chromium (...)
 * [ ] Gecko (...)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/html-aam/pull/239.html" title="Last updated on Nov 20, 2021, 7:43 PM UTC (8bf8539)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/html-aam/239/820b92f...8bf8539.html" title="Last updated on Nov 20, 2021, 7:43 PM UTC (8bf8539)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -5709,10 +5709,18 @@
   </section>
   <section class="normative">
     <h2>Accessible Name and Description Computation</h2>
-    <p>The terms <a class="termref">accessible name</a> and <a class="termref">accessible description</a> are properties provided in all <a class="termref">accessibility APIs</a>. The name of the properties may differ across APIs but they serve the same function: as a container for a short (name) or longer (description) string of text. </p>
+    <p>The terms <a class="termref">accessible name</a> and <a class="termref">accessible description</a> are properties provided in all <a class="termref">accessibility APIs</a>. The name of the properties may differ across APIs but they serve the same function: as a container for a short (name) or longer (description) string of text.</p>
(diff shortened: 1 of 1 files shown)


dd8:
May not make a difference to this PR but worth noting there are 2 kinds of whitespace involved in accname calculations and other API mappings:

1) Name whitespace which appears in the actual name/description exposed via the API (unless trimmed)

2) Syntactic whitespace defined by the host language spec. HTML defines space separated tokens which uses specific ASCII code points as separators. This is used to split to class names in `class` attributes and between IDs in IDREF attributes like `aria-labelledby` `headers` and `output for`
https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#space-separated-tokens

The two are not necessarily identical - you may want to trim additional Unicode space characters in name whitespace. Also worth noting syntactic whitespace differs by host language spec (e.g. XHTML has 4 whitespace code points (space, tab, CR, LF) and HTML5 has 5 whitespace code points (space, tab, CR, LF plus form-feed)

scottaohara:
since making this PR, there have been other updates to accName for handling whitespace.  and honestly, HTML AAM should not be defining how to handle white space in accNames separately from what the accName spec says to do.  So if there are any further clarifications needed, those should be made over in accname.

PR #271, opened 2019-12-03
Add hgroup, slot and autonomous custom element

### LINKED ISSUE #189: Add HTML elements

carmacleod:
Closes #189

***
<a href="https://pr-preview.s3.amazonaws.com/carmacleod/html-aam/pull/271.html" title="Last updated on Dec 12, 2019, 1:19 PM UTC (0a26939)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/html-aam/271/5787a05...carmacleod:0a26939.html" title="Last updated on Dec 12, 2019, 1:19 PM UTC (0a26939)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -486,7 +486,7 @@
               <th>
                 <a data-cite="HTML">autonomous custom element</a>
               </th>
-              <td class="aria">Use explicitly declared role. If none, then no corresponding role</td>
+              <td class="aria">Map to ARIA role defined by author. Otherwise, no corresponding role.</td>
               <td class="ia2"><div class="general">Use WAI-ARIA mapping</div></td>
               <td class="uia"><div class="general">Use WAI-ARIA mapping</div></td>
               <td class="atk"><div class="general">Use WAI-ARIA mapping</div></td>
@@ -1452,7 +1452,7 @@
               <th>
                 <a href="https://html.spec.whatwg.org/multipage/custom-elements.html#custom-elements-face-example">form-associated custom element</a>
               </th>
-              <td class="aria">Use explicitly declared role. If none, then no corresponding role</td>
+              <td class="aria">Map to ARIA role defined by author. Otherwise, no corresponding role.</td>
               <td class="ia2"><div class="general">Use WAI-ARIA mapping</div></td>
               <td class="uia"><div class="general">Use WAI-ARIA mapping</div></td>
               <td class="atk"><div class="general">Use WAI-ARIA mapping</div></td>
--- a/index.html
+++ b/index.html
@@ -1450,6 +1450,17 @@
               </td>
               <td class="comments"></td>
             </tr>
+            <tr tabindex="-1" id="el-form-associated-custom-element">
+              <th>
+                <a href="https://html.spec.whatwg.org/multipage/custom-elements.html#custom-elements-face-example">form-associated custom element</a>
+              </th>
+              <td class="aria">Use explicitly declared role. If none, then no corresponding role</td>
+              <td class="ia2"><div class="general">Use WAI-ARIA mapping</div></td>
+              <td class="uia"><div class="general">Use WAI-ARIA mapping</div></td>
+              <td class="atk"><div class="general">Use WAI-ARIA mapping</div></td>
+              <td class="ax"><div class="general">Use WAI-ARIA mapping</div></td>
+              <td class="comments"></td>
+            </tr>
             <tr tabindex="-1" id="el-h1-h6">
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -484,6 +484,17 @@ <h3>HTML Element Role Mappings</h3>
...
+            <tr tabindex="-1" id="el-autonomous-custom-element">
+              <th>
+                <a data-cite="HTML">autonomous custom element</a>
+              </th>
+              <td class="aria">Use explicitly declared role. If none, then no corresponding role</td>
+              <td class="ia2"><div class="general">Use WAI-ARIA mapping</div></td>
+              <td class="uia"><div class="general">Use WAI-ARIA mapping</div></td>
+              <td class="atk"><div class="general">Use WAI-ARIA mapping</div></td>
+              <td class="ax"><div class="general">Use WAI-ARIA mapping</div></td>
+              <td class="comments"></td>
```
scottaohara, line 494: We should also add "[form-associated custom element](https://html.spec.whatwg.org/multipage/custom-elements.html#custom-elements-face-example)" as either a comment here of a type of autonomous custom element, or as its own entry in the table.

carmacleod: Ah, I see what you meant earlier.
I originally didn't think it was a separate thing (thought it was essentially just a glorified boolean ;).
Same mappings as regular autonomous-c-e, though, right?
Cool that form-associated-c-e can be labelled by ancestor label element.
Maybe that's a reason to give it its own entry, in case that uses some special mapping some day?
If you prefer all in a-c-e, let me know.

index.html
```diff
@@ -1450,6 +1450,17 @@ <h3>HTML Element Role Mappings</h3>
               </td>
               <td class="comments"></td>
             </tr>
+            <tr tabindex="-1" id="el-form-associated-custom-element">
+              <th>
+                <a href="https://html.spec.whatwg.org/multipage/custom-elements.html#custom-elements-face-example">form-associated custom element</a>
+              </th>
+              <td class="aria">Use explicitly declared role. If none, then no corresponding role</td>
```
scottaohara: i wonder what a form associated custom element would even be if it had no role?

carmacleod: heh, yeah. sounds like "author error". so... "undefined".
Does "If none, then no corresponding role" cover that?

carmacleod:
Hi @marcoscaceres. Can you please look into this respec error?
I'm trying to get this PR in because I need it for an [HTML PR](https://github.com/whatwg/html/pull/5077) I am working on.
Thanks!

marcoscaceres:
Marked as non substantive for IPR from ash-nazg.

marcoscaceres:
Hi @carmacleod, I think I fixed the ReSpec issue.... 

PS: we really appreciate your participation and contribution to this work. Would you/IBM be open to joining the WG? We'd really like to have you join us.

index.html
```diff
@@ -482,6 +482,17 @@ <h3>HTML Element Role Mappings</h3>
               </td>
               <td class="comments"></td>
             </tr>
+            <tr tabindex="-1" id="el-autonomous-custom-element">
+              <th>
+                <a data-cite="HTML">autonomous custom element</a>
+              </th>
+              <td class="aria">Use explicitly declared role. If none, then no corresponding role</td>
```
scottaohara: I'm thinking "Map to ARIA role defined by author. Otherwise, no corresponding role."

Where "no corresponding role" will need to be changed in the future to say `role=generic`.

index.html
scottaohara: Similar to my previous comment:
"Map to ARIA role defined by author. Otherwise, no corresponding role."

carmacleod:
Thanks for the fix and the invitation, @marcoscaceres ! The process has been initiated. :)
I really only have time for ARIA in HTML and HTML AAM, though. (Along with ARIA and APG).

scottaohara (approved):

marcoscaceres:
> Thanks for the fix and the invitation, @marcoscaceres ! The process has been initiated. :)
I really only have time for ARIA in HTML and HTML AAM, though. (Along with ARIA and APG).

Understood. Definitely won't drag you into any of the other specs :)

index.html
```diff
@@ -482,6 +482,17 @@ <h3>HTML Element Role Mappings</h3>
               </td>
               <td class="comments"></td>
             </tr>
+            <tr tabindex="-1" id="el-autonomous-custom-element">
+              <th>
+                <a data-cite="HTML">autonomous custom element</a>
+              </th>
+              <td class="aria">Map to ARIA role defined by author. Otherwise, no corresponding role.</td>
```
stevefaulkner: This seems a bit woolly to me. something more like "Map to the conforming ARIA role value assigned using the role attribute, if present." Otherwise...

not saying the above be used verbatim, but think that it needs to be made clearer.

carmacleod: Here are 2 other possibilities:
1. Use the conforming ARIA role assigned using the role attribute, if present. Otherwise, no corresponding role.
2. If the author assigned a conforming ARIA role using the role attribute, map to that role. Otherwise, no corresponding role.

Personal preference for the 2nd one, although it's a bit longer. Your suggested wording would be fine, too. Which of these seems clearest?

PS: Good point about "conforming". Don't want folks to try putting `role="listbox"` on a custom element based on ``. ;)

carmacleod: I went ahead and committed the 2nd sentence, above. Of course, that can be changed if other wording is preferred.

stevefaulkner (approved):

Issue #259, opened 2019-10-18
Update ARIA mapping to remove instances of "?" for HTML attributes


scottaohara:
Related to issue #97, there are still instances of "?" in the document related to ARIA mappings.  

specifically:

- [ ] `contenteditable` (HTML elements) 
- [x] `span` (col; colgroup)
- [x]  `title` (HTML elements)

joanmarie:
Regarding `contenteditable`: Should it trump the generic role? See https://github.com/w3c/aria/discussions/1449.

scottaohara:
yes.  well.  for your specific example, yes.

for more complex `<button contenteditable>...</button>`, i honestly am not sure what should happen there.

but for generic divs, those (at least last time i checked) get exposed as textboxes.

joanmarie:
At least in the case of Chromium, `data:text/html,<div contenteditable>foo</div>` has a computed role of `generic` when I look in the dev tools.

As a result, my (local; uncommitted) implementation in Chromium to do the ARIA 1.2 name prohibiting makes `data:text/html,<div contenteditable aria-label="name">foo</div>` not have a name. Ditto for a focusable, but not contenteditable, `div`. While that's not the best authoring, prohibiting exposure of the `aria-label` value on these interactive things also seems bad. Furthermore, I can see us getting pushback from the browser vendors.

I think one way to solve this problem is for your HTML-AAM to cause `contenteditable` (and `tabindex` >= 0) on generic elements map to a non-generic role. I'm open to other ideas. 

Note that as James verified in https://github.com/w3c/aria/discussions/1449:

> https://w3c.github.io/aria/#document-handling_author-errors_states-properties states "User Agents MUST NOT expose an ARIA attribute to the platform accessibilty API when the element's computed role prohibits use of the attribute."

So for ARIA 1.2 to exit CR, I believe we're going to need a solution ... somewhere.

scottaohara:
Took another gander at [my test page](https://scottaohara.github.io/tests/attr-html/contenteditable.html), and yeh, don't recall if things changed, or if i was misremembering this to be more consistent - but only webkit treats `contenteditable` on a `div` as if it were a `textbox`, exposing the `AXTextArea` at the platform level.  But on other elements that expose roles (though not all) the base role is retained.  e.g., `button` and `main` continue to be exposed.  but the`aside`'s `complementary` role is dropped.

[...]

PR #324, opened 2021-04-08
updates to contenteditable and tabindex

scottaohara:
remap ‘no corresponding role’ elements (i.e., `role=generic` or inline generic-ish elements) to `role=group` if they have `contenteditable` or `tabindex` set.

related to #259

Review links (because unfortunately the pr-preview Preview and Diff links below are useless):
https://raw.githack.com/w3c/html-aam/contenteditable-tabindex-updates/index.html#att-contenteditable
https://raw.githack.com/w3c/html-aam/contenteditable-tabindex-updates/index.html#att-tabindex

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/html-aam/pull/324.html" title="Last updated on Apr 22, 2021, 2:49 PM UTC (5235d98)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/html-aam/324/f1b695c...5235d98.html" title="Last updated on Apr 22, 2021, 2:49 PM UTC (5235d98)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -4221,11 +4221,18 @@
               <td class="elements">
                 <a data-cite="html/interaction.html#attr-contenteditable">HTML elements</a>
               </td>
-              <td class="aria">?</td>
+              <td class="aria">
+                <p class="general">
+                  If set to an element which already exposes a corresponding role, then no specific ARIA mapping.
+                </p>
+                <p>
+                  If set to an element with no corresponding role, then map to the <a class="core-mapping" href="#role-map-group">`group`</a> role.
+                </p>
+              </td>
               <td class="ia2">
                 <div class="states">
                   <span class="type">States:</span>
-                  `IA2_STATE_EDITABLE` on this and every nested text accessible object
+                  `IA2_STATE_EDITABLE` and `IA2_STATE_MULTI_LINE` on this and every nested text accessible object
                 </div>
                 <div class="interfaces">
                   <span class="type">Interfaces:</span>
@@ -4240,7 +4247,7 @@
               <td class="atk">
                 <div class="states">
                   <span class="type">States:</span>
-                  `ATK_STATE_EDITABLE` on this and every nested text accessible object.
+                  `ATK_STATE_EDITABLE` and `STATE_MULTI_LINE` on this and every nested text accessible object.
                 </div>
                 <div class="interfaces">
                   <span class="type">Interfaces:</span>
@@ -5689,6 +5696,9 @@
               </td>
               <td class="aria">
                 <a class="core-mapping" href="#focus_state_event_table">See Focus States and Events Table</a>
+                <p>
+                  If set to an element with no corresponding role, then map to the <a class="core-mapping" href="#role-map-group">`group`</a> role.
+                </p>
               </td>
               <td class="ia2"><div class="general">Use WAI-ARIA mapping</div></td>
               <td class="uia"><div class="general">Use WAI-ARIA mapping</div></td>
--- a/index.html
+++ b/index.html
@@ -244,7 +244,7 @@
       		<thead>
       			<tr>
       				<th>Element</th>
-      				<th>[[wai-aria-1.1]]</th>
+      				<th>[[[wai-aria-1.1]]]</th>
       				<th><a href="https://msdn.microsoft.com/en-us/library/dd373608%28v=VS.85%29.aspx">MSAA</a> + <a href="http://accessibility.linuxfoundation.org/a11yspecs/ia2/docs/html/">IAccessible2</a></th>
               <th><a href="https://msdn.microsoft.com/en-us/library/ms726297%28v=VS.85%29.aspx">UIA</a></th>
       				<th><a href="https://developer.gnome.org/atk/stable/">ATK</a></th>
@@ -3823,7 +3823,7 @@
             <tr>
               <th>Attribute</th>
               <th>Element(s)</th>
-              <th>[[[WAI-ARIA]]]</th>
+              <th>[[[wai-aria-1.1]]]</th>
               <th><a href="https://msdn.microsoft.com/en-us/library/dd373608%28v=VS.85%29.aspx">MSAA</a> + <a href="http://accessibility.linuxfoundation.org/a11yspecs/ia2/docs/html/">IAccessible2</a></th>
               <th><a href="https://msdn.microsoft.com/en-us/library/ms726297%28v=VS.85%29.aspx">UIA</a></th>
               <th><a href="https://developer.gnome.org/atk/stable/">ATK</a></th>
@@ -4222,12 +4222,10 @@
                 <a data-cite="html/interaction.html#attr-contenteditable">HTML elements</a>
               </td>
               <td class="aria">
-                <p class="general">
-                  If set to an element which already exposes a corresponding role, then no specific ARIA mapping.
-                </p>
                 <p>
-                  If set to an element with no corresponding role, then map to the <a class="core-mapping" href="#role-map-group">`group`</a> role.
+                  If specified on an element with no implicit WAI-ARIA role, or if the element maps to `role="generic"`, then map to the <a class="core-mapping" href="#role-map-group">`group`</a> role.
                 </p>
+                <p>Otherwise, no specific WAI-ARIA mapping.</p>
               </td>
               <td class="ia2">
                 <div class="states">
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -4221,11 +4221,18 @@ <h3>HTML Attribute State and Property Mappings</h3>
               <td class="elements">
                 <a data-cite="html/interaction.html#attr-contenteditable">HTML elements</a>
               </td>
-              <td class="aria">?</td>
+              <td class="aria">
+                <p class="general">
+                  If set to an element which already exposes a corresponding role, then no specific ARIA mapping.
```
carmacleod:
```suggestion
                  If set on an element which already exposes a role, then no specific ARIA mapping.
```

Notes:
- I find the "set to" wording a little odd because I am expecting a _value_ to follow the word "to" (so I changed it to "set on")
- Alternatively, did you want to say, 'If set to "true" or "" on an element...'
- Not quite sure whether "corresponding" is needed?

index.html
```diff
@@ -4221,11 +4221,18 @@ <h3>HTML Attribute State and Property Mappings</h3>
...
+                </p>
+                <p>
+                  If set to an element with no corresponding role, then map to the <a class="core-mapping" href="#role-map-group">`group`</a> role.
```
carmacleod:
```suggestion
                  If set on an element with no role, then map to the <a class="core-mapping" href="#role-map-group">`group`</a> role.
```

Same Notes as above.

index.html
```diff
@@ -4221,11 +4221,18 @@ <h3>HTML Attribute State and Property Mappings</h3>
...
+                </p>
+              </td>
               <td class="ia2">
                 <div class="states">
                   <span class="type">States:</span>
-                  `IA2_STATE_EDITABLE` on this and every nested text accessible object
+                  `IA2_STATE_EDITABLE` and `IA2_STATE_MULTI_LINE` on this and every nested text accessible object
```
carmacleod: Add "." for consistency.
```suggestion
                  `IA2_STATE_EDITABLE` and `IA2_STATE_MULTI_LINE` on this and every nested text accessible object.
```

index.html
```diff
@@ -4240,7 +4247,7 @@ <h3>HTML Attribute State and Property Mappings</h3>
               <td class="atk">
                 <div class="states">
                   <span class="type">States:</span>
-                  `ATK_STATE_EDITABLE` on this and every nested text accessible object.
+                  `ATK_STATE_EDITABLE` and `STATE_MULTI_LINE` on this and every nested text accessible object.
```
carmacleod: Isn't it: ATK_STATE_MULTI_LINE ?
```suggestion
                  `ATK_STATE_EDITABLE` and `ATK_STATE_MULTI_LINE` on this and every nested text accessible object.
```

carmacleod (commented):
Thanks, @scottaohara !

Just a few notes, below, plus a question: Is it clear what "text accessible object" means?

index.html
```diff
@@ -5689,6 +5696,9 @@ <h3>HTML Attribute State and Property Mappings</h3>
               </td>
               <td class="aria">
                 <a class="core-mapping" href="#focus_state_event_table">See Focus States and Events Table</a>
+                <p>
+                  If set to an element with no corresponding role, then map to the <a class="core-mapping" href="#role-map-group">`group`</a> role.
```
carmacleod: Similar notes to above... and one more question:
Did you want to say 'If set to a value >= "0" on an element with no role, then...' ?
(because this includes "-1" as-is - not sure if you wanted to include that)

```suggestion
                  If set on an element with no role, then map to the <a class="core-mapping" href="#role-map-group">`group`</a> role.
```

index.html
scottaohara: - "set on" is fine by me.
- yeah, that should be more clear... while "true" and "" are the same...but since "false" matters here, it should really say "if contenteditable is true"
- i did mean to use "corresponding" per its use in other parts of the spec.  we can't just say "which already exposes a role", because "generic" is a role, and that's what we're trying to mitigate against here.  So this really should be something like "if set on an element which exposes an implicit aria role, other than generic, then no specific ARIA mapping." - to indicate that `contenteditable` has no other ARIA mapping itself

index.html
carmacleod: I downloaded, un-xzipped, and un-tarred the latest ATK (2.36.0 - hope that's the latest) and found `ATK_STATE_` prefixed constants in file atkstate.h.

But there seems to also be some pythony thing at https://developer.gnome.org/pygtk/stable/atk-constants.html
which uses an `atk` object for the constants and doesn't bother with the `ATK_` prefix.

It looks like CORE-AAM doesn't use the ATK_ prefix on any of the ROLE, STATE, or RELATION constants, so I'm guessing that's the preferred naming nowadays? @joanmarie ?

scottaohara: yes, i was looking at CORE AAM when putting that in there.  granted, that may be unnecessary, so would love @joanmarie's feedback on that.  but also, would like to know whether i should be using the prefixed or unprefixed versions in this spec

joanmarie: Others from the past thought that the AAM should not have the prefix because then it could apply to both AT-SPI2 and ATK. Since implementors are implementing ATK, I would have preferred ATK to be there because it's technically more correct. But.... consensus. ;)

From a practical point of view, implementors know that they are implementing ATK and that they therefore need to actually use the correct form of the constant which has the prefix. So they will implement `ATK_STATE_EDITABLE` (which, you know, compiles) even if a spec says only `STATE_EDITABLE` (which will fail to compile).

And for the most part, ATK and AT-SPI2 are similar. Though every once in a while there's an inconsistency (e.g. it's `ATK_ROLE_STATUSBAR`, but `ATSPI_ROLE_STATUS_BAR`. Oops!). There are also differences in certain methods, e.g. `atk_table_ref_at()`, versus `atspi_table_get_accessible_at()`).

As for the python bindings for ATK.... Let's cross that bridge when someone creates a written-in-python browser which they want to make accessible. ;)

As for which you should use @scottaohara, I don't have a strong feeling either way (see first paragraph above). What's more important (of course) is to pick one and be consistent.

joanmarie:
Operating under the assumption that this is going to land, I'm working on making the change in Chromium. (This is independent of the name prohibition code changes btw.) Running Chromium's tests to see what broke led to me finding another thing we need to handle somewhere:

The `group` role *does* allow naming BUT from author; not from content. So some things which were generic lost their names as a consequence of becoming a group with no author-provided name. Oops!

Thoughts?

scottaohara:
in the context of contenteditable, if not specifically named by an author, it exposing no name makes sense to me.  since large areas of content can have a contenteditable parent, we would not want all that content contributing to the parent's accessible name.

Similar with tabindex=0 on a scrollable div of content.  

what are some of the things you are finding that lost names that I might not be considering here, that we might want to maintain?

joanmarie:
> what are some of the things you are finding that lost names that I might not be considering here, that we might want to maintain?

One test that broke had this: `<div tabindex=0>Focusable div</div>`
* Expected results for blink (reminder: I've not done the name-prohibited stuff): `genericContainer name='Focusable div'`
* Actual results for blink: `group`

If a user tabs to that focusable div, they should get a name presented to them, right?

scottaohara:
@joanmarie yeh that's a reduced example of these https://codepen.io/scottohara/pen/bGBaYEQ

where i think it'd be quite problematic for all that content to be exposed as the name.  I don't think `<div tabindex=0>` should get name from contents.

jnurthen:
IMO this is something that should fit into the "handling author errors" of the ARIA spec. So if something gets focus and has no accessible name then author agents MAY correct by doing (a) (b) (c) etc.

joanmarie:
@scottaohara and @jnurthen I'm away on "vacation" next week. But it would be great to have consensus on this so that we can proceed doing/getting our 1.2 implementations. :smile:

index.html
scottaohara: updated to ATK_STATE_MULTI_LINE
resolving

aleventhal:
I think role=group may be too simple of a repair.
For a contenteditable, I would suggest role=textbox is more appropriate. The browser can even try to set aria-multiline appropriately based on whether the field's CSS limits rendering to a single line.
We could even map to role=button if there is a click handler.

scottaohara:
@aleventhal have you reviewed the discussion on this in #259?

PR #329, opened 2021-04-25
dialog open attribute mapping

### LINKED ISSUE #265: Mapping of "open" on dialog seems wrong
According to https://html.spec.whatwg.org/multipage/interactive-elements.html#the-dialog-element:

> The open attribute is a boolean attribute. When specified, it indicates that the dialog element is active and that the user can interact with it.
 > A dialog element without an open attribute specified should not be shown to the user.

Dialogs -- at least on my platform -- are NOT things which can be expanded or collapsed. They can be shown or hidden/destroyed. When they're showing, they should be included in the accessibility tree; when they are hidden/destroyed, they should not be in the accessibility tree.

My *guess* is that this is the case for all the platforms.

Aside: The mapping for "open" on details seems right. Maybe this is just a copy and paste error for dialog?

scottaohara:
Closes #265

@joanmarie finally took a stab at revising the incorrect `open` mappings for the `dialog` element.  Not sure if the `aria-modal` references need to be in there. But wanted to call out that `open` can represent a non-modal or modal dialog depending on the method that it is specified on the element.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/html-aam/pull/329.html" title="Last updated on Apr 26, 2021, 12:56 PM UTC (e638b14)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/html-aam/329/08ac91c...e638b14.html" title="Last updated on Apr 26, 2021, 12:56 PM UTC (e638b14)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1067,7 +1067,9 @@
               <td class="uia"><div class="general">Use WAI-ARIA mapping</div></td>
               <td class="atk"><div class="general">Use WAI-ARIA mapping</div></td>
               <td class="ax"><div class="general">Use WAI-ARIA mapping</div></td>
-              <td class="comments"></td>
+              <td class="comments">
+                See also the `dialog` element's <a href="#att-open-dialog">`open`</a> attribute.
+              </td>
             </tr>
             <tr tabindex="-1" id="el-div">
               <th>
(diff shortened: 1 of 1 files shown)


joanmarie (approved):
I'm fine with this as-is because it corrects the problem (i.e. removal of the expanded-related mapping).

That said, some observations:
* `open` appears to be a boolean. The mappings you describe appear to just be for `true`. And if I'm reading the HTML spec correctly, user agents have some flexibility regarding how to implement `false`/absence means "should not be shown to the user". So does it make sense to also cover the `false` part? ("nope" is a valid response; just thinking out loud)
* How is `display: block` relevant to the accessibility API mappings? (again, maybe it is; simply asking)

scottaohara:
@joanmarie made updates to hopefully clarify.

joanmarie:
Thanks! LGTM.

PR #334, opened 2021-05-29
Fill in ARIA rows - issue 259

scottaohara:
partially resolves 259

* `<col>` and `<colgroup>` `span` attribute ARIA relationship
* `title` attribute ARIA relationship

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/html-aam/pull/334.html" title="Last updated on Jul 20, 2021, 3:06 PM UTC (50c8701)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/html-aam/334/b9ef143...50c8701.html" title="Last updated on Jul 20, 2021, 3:06 PM UTC (50c8701)">Diff</a>


scottaohara:
@joanmarie would appreciate your thoughts on these changes and if i'm not thinking of something - specifically for `title`

joanmarie:
Regarding colgroup:

1. I agree there's no corresponding ARIA role. :smile: 
2. Aside: While this PR doesn't touch the mac mapping, I think it's exposed. Just viewed a table with a `colgroup` in Safari and used Accessibility Inspector to see what was there. The table had children for each row, followed by children for each column, followed by an AXGroup. Might be worth tackling in a separate PR after verifying the mappings.

Regarding title:
1. I'm not opposed to the wording, but I'm not jazzed by it either. What are you trying to convey specifically?
2. Related: The "otherwise then accessible description" and "AXHelp: <value>", which you're not changing in this PR, is of course only true of nothing else trumped it (e.g. because both `aria-label` and `aria-description` were both present).

If it were me, I might instead change the ARIA "mapping" (and maybe the platform mappings too?) to something boiling down to "It's complicated. See AccName."

scottaohara:
thanks @joanmarie and understood on the less than jazzy feeling towards the wording.  I was basically trying to keep in line with how the 'aria' mapping columns had been used and associate the attribute with the corresponding attribute/concept that is in core aam.  so when title provides the name, that's most like what aria-label does as title would specify the accName, but not by referencing another element like aria-labelledby.  Similar with when it provides a description - it just does that directly w/out referencing another element.

But with that said, I think I like your idea of saying for all rows that the title will either contribute to the accName or the description, and to look to the name/desc .

joanmarie:
> But with that said, I think I like your idea of saying for all rows that the title will either contribute to the accName or the description, and to look to the name/desc .

Nit: s/will/might/ :grinning: #ItsComplicated

scottaohara:
@joanmarie :D

do you have spec-approved verbage i can use for "there be dragons. good luck yo!"

if so, i will add it just for you :)

joanmarie:
(NOTE: Edited to tweak the proposed comments text)

So patches welcome, eh? :grinning: Ok, that's fair.

Dunno about "spec-approved", but taking a stab first at the "Comments" row for "title":

> The [Accessible Name and Description Computation](https://w3c.github.io/html-aam/#accessible-name-and-description-computation) section specifies if the `title` will be mapped and, if so, through what WAI-ARIA property.

And if that (or something like that) is the "Comments", then there is a WAI-ARIA property. So I'd turn each platform row to:

> Use WAI-ARIA mapping

Which means we just need something for the WAI-ARIA mapping, which is the harder part. What about:

> accessible name, or accessible description, or not mapped (see Comments)

scottaohara:
Thanks @joanmarie.  I've updated per your feedback.  

backing up a bit re: the AX mapping of `col` and `colgroup`.   I noticed that same extra 'group' in the table regardless of whether there was a `colgroup` specified or not.  It only appears to have child cells when/if column headers are present.  If there are no column headers, then the group is empty and doesn't appear to reflect the colgroup/col markup.   

Made a demo: https://codepen.io/scottohara/full/rNmNbaL

stevefaulkner (approved):

PR #350, opened 2021-10-29
aside mapping revisions

### LINKED ISSUE #86: Suggest <aside> should map to complementary with restrictions
Like the `header` and `footer` and `section` elements, the `aside` element can be used liberally within a HTML document. This liberal use can lead to the semantics of the element becoming noise to aural UI users. What we did for the `header/footer` elements is to restrict the expression of the role semantics based on their nesting within other significant elements in the DOM. For the `section` element we restricted its expression based on the presence of an accessible name. I suggest we should do the same/similar for the `aside` element.
[grep of aside usage](https://dl.dropboxusercontent.com/u/377471/aside/aside-grep.html) - '<aside|</aside>' in *.txt: 9648 matches in 1376 files. 25221 files searched.

**note:** The data is derived from [webdevdata.org](http://webdevdata.org/) latest dump (2016)

**Some examples from above of pages with high usage up to 100+**
The pages have been saved with their aside elements indicated visually:
- [Bola.com](https://dl.dropboxusercontent.com/u/377471/aside/Bola.com.html)
- [efectococuyo.com](https://dl.dropboxusercontent.com/u/377471/aside/Efecto%20Cocuyo%20-%20Periodismo%20que%20Ilumina.html)
- [Grafica.info.com](https://dl.dropboxusercontent.com/u/377471/aside/gr%C3%A0ffica.info.html)
- [Infirmiers.com](https://dl.dropboxusercontent.com/u/377471/aside/Infirmiers.com.html)
- [pz-news.de](https://dl.dropboxusercontent.com/u/377471/aside/PZ-news.de.html)
- [splitsider.com](https://dl.dropboxusercontent.com/u/377471/aside/Splitsider.html)

/cc @cookiecrook @asurkov @w3c/kiss

scottaohara:
closes #86

Treat an `aside` element similarly to `header`, `footer` and `section`.

map `aside` to `role=complementary` if:
* scoped to `main` or `body` elements
* or given accessible name if scoped to sectioning content or root elements

otherwise, map to `role=generic`

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/html-aam/pull/350.html" title="Last updated on Nov 7, 2021, 6:41 PM UTC (ab6b093)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/html-aam/350/538a078...ab6b093.html" title="Last updated on Nov 7, 2021, 6:41 PM UTC (ab6b093)">Diff</a>

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/html-aam/pull/350.html" title="Last updated on Apr 3, 2022, 4:30 PM UTC (e429a37)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/html-aam/350/13721fb...e429a37.html" title="Last updated on Apr 3, 2022, 4:30 PM UTC (e429a37)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -433,7 +433,9 @@
             </tr>
             <tr tabindex="-1" id="el-aside">
               <th>
-                <a data-cite="HTML">`aside`</a> (scoped to a sectioning content element, or a sectioning root element other than `body`)
+                <a data-cite="HTML">`aside`</a> 
+                (scoped to a <a data-cite="HTML/dom.html#sectioning-content">sectioning content</a> element, 
+                or a <a data-cite="HTML/sections.html#sectioning-root">sectioning root</a> element other than `body`)
               </th>
               <td class="aria">
                 <a class="core-mapping" href="#role-map-complementary">`complementary`</a> role if the <a>`aside`</a> element has an <a>accessible name</a>. 
--- a/index.html
+++ b/index.html
@@ -458,18 +458,29 @@
               <td class="ax"><div class="general">Use WAI-ARIA mapping</div></td>
               <td class="comments"></td>
             </tr>
-            <tr tabindex="-1" id="el-aside">
+            <tr tabindex="-1" id="el-aside-ancestorbodymain">
               <th>
-                <a data-cite="HTML">`aside`</a>
+                <a data-cite="HTML">`aside`</a> (scoped to the `body` or `main` element)
               </th>
               <td class="aria">
                 <a class="core-mapping" href="#role-map-complementary">`complementary`</a> role
               </td>
               <td class="ia2"><div class="general">Use WAI-ARIA mapping</div></td>
-              <td class="uia">
-                <div class="general">Use WAI-ARIA mapping</div>
-                <div class="ctrltype"><span class="type">Localized Control Type:</span> `"aside"`</div>
+              <td class="uia"><div class="general">Use WAI-ARIA mapping</div></td>
+              <td class="atk"><div class="general">Use WAI-ARIA mapping</div></td>
+              <td class="ax"><div class="general">Use WAI-ARIA mapping</div></td>
+              <td class="comments"></td>
+            </tr>
+            <tr tabindex="-1" id="el-aside">
+              <th>
+                <a data-cite="HTML">`aside`</a> (scoped to a sectioning content element, or a sectioning root element other than `body`)
+              </th>
+              <td class="aria">
+                <a class="core-mapping" href="#role-map-complementary">`complementary`</a> role if the <a>`aside`</a> element has an <a>accessible name</a>. 
+                Otherwise, <a class="core-mapping" href="#role-map-generic">`generic`</a> role.
               </td>
+              <td class="ia2"><div class="general">Use WAI-ARIA mapping</div></td>
+              <td class="uia"><div class="general">Use WAI-ARIA mapping</div></td>
               <td class="atk"><div class="general">Use WAI-ARIA mapping</div></td>
               <td class="ax"><div class="general">Use WAI-ARIA mapping</div></td>
               <td class="comments"></td>


aleventhal (commented):
Is there already consensus that it's a good idea?

I don't see any problems from an implementation point of view.

cookiecrook:
After a brief internal discussion, we think it’s a reasonable proposal, with a relatively simple implementation. Assuming no objections of merit arise, of course.

scottaohara:
thanks @cookiecrook, @aleventhal 

the tldr; intent of this update is to mitigate against the guidance from ARIA / the ARIA Authoring Practices (and conformance checkers that reference APG) that indicate that a `complementary` landmark should generally be a top-level landmark, while the HTML specification indicates that use of an `<aside>` element can be far more permissive.  

This change would allow for the `<aside>` to better handle both sets of authoring use, and then authors that specifically want to make sure an `<aside>` is exposed as a complementary landmark can still do so by giving it an accessible name, regardless of where it lives in the DOM.  Otherwise, it allows authors to continue to use it without unknowingly overpopulating the number of complementary landmarks in the document.

scottaohara:
[test case created](https://scottaohara.github.io/testing/aside/)

- [webkit bug filed](https://bugs.webkit.org/show_bug.cgi?id=235065)
- [chromium bug filed](https://bugs.chromium.org/p/chromium/issues/detail?id=1286298#makechanges)
- [firefox bug filed](https://bugzilla.mozilla.org/show_bug.cgi?id=1749505)

spectranaut:
ok @scottaohara I'm trying to implement this, and using your useful test case, and I ran into a bug in the conditional mapping of footer and header: https://bugs.chromium.org/p/chromium/issues/detail?id=1303917

I'm commenting here, though, because I think there is some lack of clarity in the spec. I couldn't find a definition of "sectioning element" and "sectioning root element"... are those phrases defined in some other spec? Or are they open to interpretation? If a sectioning element means a `<section>` (or `<aside>` or whatever the list of "sectioning elements" is) regardless of the ARIA role, for example, then your test case is wrong and the chrome implementation is correct.

Also is "scoped" defined somewhere? Regarding: "footer (scoped to the `body` or `main` element)". Trying to read this in order to implement it was difficult, is what I'm getting at. If "scoped" means "has a parent which is...", then a footer in some html might fall under both categories. For example, an `<footer>` inside an `<article>` inside of `<main>`. Could they be worded in a way that more obviously makes it mutually exclusive scenarios?

If neither of these things are defined anywhere, maybe this is the PR to add them? :)

scottaohara:
@spectranaut [sectioning contnet](https://html.spec.whatwg.org/multipage/dom.html#sectioning-content) and [sectioning root](https://html.spec.whatwg.org/multipage/sections.html#sectioning-root) are defined in HTML.

But to your point, yes, I think you're right and we need to define 'scoped' here to mitigate this question and help simplify `header`, `footer` and now `aside` which all reference this.

and to be honest, we could probably even simplify that definition, as truly it's not just those elements... but if say a `div role=region` were to interrupt the descendant relationship of the `body` and `aside`, at the element level that would be _ok_, because a `div` is `generic` and doesn't fall into one of those categories.  But by an author making it a `role=region`, that `aside` should no longer be a `role=complementary`.

EDIT: apologies. I realized when looking at this that it sure would have helped if I remembered to link to the HTML spec per those terms.  Those links have been added in.

spectranaut:
Awesome thanks for adding the links, that helps!

I'm not sure what you are suggesting exactly for simpler categories, but I was just thinking something like:
1.  Aside (scoped to a sectioning content element, or a sectioning root element other than `body`)
2.  Aside (not scoped to a sectioning content element, or a sectioning root element other than `body`)

I still don't know if "scoped" is the right word or where the definition for it should go, or if it is self-evident to people other than me?

scottaohara:
Since it's not an officially defined term as far as i can tell, i'm not opposed to changing it to another word(s) if that makes the most sense.  There are other similar terms in ARIA and HTML that sorta fit this concept....

anyway, sent you an email to try and discuss further outside of back and forth issue comments.  if that doesn't work for you though, happy to keep at this until we get this sorted.

stevefaulkner (approved):

cookiecrook (commented):
I think this needs an update to the following WPT test:

https://wpt.fyi/results/html-aam/roles-contextual.html?label=master&label=experimental&aligned&q=label%3Aaccessibility

https://github.com/web-platform-tests/wpt/blob/master/html-aam/roles-contextual.html

PR #354, opened 2021-11-03
revise mapping tables for svg and math

### LINKED ISSUE #344: Mapping of svg/math tags in non-SVG/non-MathML namespaces
cc @asurkov 

The HTML-aam spec defines "how user agents map HTML elements" which I interpret it as elements in the HTML namespace ( see https://html.spec.whatwg.org/multipage/infrastructure.html#html-elements ). However, it also describes mapping for the `<svg>` and `<math>` elements which are *not* HTML elements, even if they are described in HTML5 (note that they are respectively put in the SVG and MathML namespaces by the parser).

Mapping for `<svg>` is actually already defined in svg-aam and it was suggested that html-aam cross-references it in https://github.com/w3c/html-aam/issues/43#issuecomment-253109071

Mapping for `<math>` is not defined in mathml-aam but it looks like this is the place where it should be ( see https://github.com/w3c/mathml-aam/issues/9 ) 

If you want to keep `<svg>` and `<math>` in the HTML-aam spec, I believe you should be explicit about namespaces to avoid confusion. For example, I noticed that Chromium seems to only check the local name to map tag `math` to role `math`, which means it currently performs the mapping for non-MathML `<math>` tag too: https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/modules/accessibility/ax_node_object.cc;l=1005;drc=4b13e65c914599d4be74a9f4db257fa1218b865b

scottaohara:
@fred-wang @joanmarie, draft update to close #344

Looks like math aam still has https://github.com/w3c/mathml-aam/issues/9 open, and `math` isn't in the mapping table yet?   Would you rather the spec link to MathML Core, or have this spec update after issue 9 is closed?

I have manual links in for mathml aam at this time, as i get respec errors when i try to cite the spec using the specification short name. e.g., would have expected to be able to do `[[[mathml-aam-1.0]]]` but I get a 'bad reference' error.

Any wording or link updates/recommendations are welcome.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/html-aam/pull/354.html" title="Last updated on Nov 3, 2021, 1:43 PM UTC (7c6a30f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/html-aam/354/12414c5...7c6a30f.html" title="Last updated on Nov 3, 2021, 1:43 PM UTC (7c6a30f)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -166,7 +166,7 @@
       <ul>
         <li><abbr title="Microsoft Active Accessibility">MSAA</abbr> with <cite><a href="https://wiki.linuxfoundation.org/accessibility/iaccessible2/start">IAccessible2 1.3</a></cite> [[IAccessible2]]</li>
         <li><cite><a href="https://msdn.microsoft.com/en-us/library/ee684013%28VS.85%29.aspx">User Interface Automation</a></cite> [[UI-AUTOMATION]]</li>
-        <li><cite>Linux/GNOME <a href="https://gnome.pages.gitlab.gnome.org/atk/">ATK - Accessibility Toolkit</a></cite> [[ATK]] and <cite><a href="https://gitlab.gnome.org/GNOME/at-spi2-core/">Assistive Technology Service Provider Interface</a></cite> [[AT-SPI]], referred to hereafter as "ATK/AT-SPI"</li>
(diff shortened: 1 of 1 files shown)


fred-wang:
@scottaohara Thanks for looking into this. I think I'll check with @asurkov to make progress on  w3c/mathml-aam#9

fred-wang:
> @fred-wang @joanmarie, draft update to close #344
> 
> Looks like math aam still has [w3c/mathml-aam#9](https://github.com/w3c/mathml-aam/issues/9) open, and `math` isn't in the mapping table yet? Would you rather the spec link to MathML Core, or have this spec update after issue 9 is closed?

@scottaohara `math` is now in the MathML AAM spec: https://w3c.github.io/mathml-aam/#mathml-element-mappings

scottaohara:
@stevefaulkner good with the linking language, before I merge this?

stevefaulkner (approved):

PR #396, opened 2022-04-15
update / expand input value mappings

### LINKED ISSUE #314: value should map to aria-valuenow for input[type=range]
Currently value is stated as not mapped for an input element.

scottaohara:
This PR closes #314 and clarifies the `value` attribute mappings for other input types.
note: the `value` attribute only represents *the initial value* for the input elements that expose the attribute's value.  If a user updates a form control the `value` attribute's value does not mutate to reflect the updated entry.  The updates I provided for these inputs attempts to specifically call this out.  if this is not clear, please suggest alternate wording.

Note that the `input type=color` `value` is far more complex than the other inputs and I would appreciate specific feedback on what would need to be updated here, if anything.

Additionally, the `input type=color` seems to make a strong case for allowing `aria-valuetext` on a `button` element – where the button should be named via a `label` or an `aria-label/labelledby`, but still needs to expose it's displayed value (subtree of the button element).

**NOTE:** due to the still broken table/details/summary script with respec, to view the updated table in the preview, you can add this style via devtools:
```
#html-attribute-state-and-property-mappings .table-container {
  display: block !important;
}
```
or find the `.table-container` that has the `display: none` on it, and turn that off.

fun...

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/html-aam/pull/396.html" title="Last updated on Oct 26, 2022, 1:25 PM UTC (0e75d89)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/html-aam/396/a5ada74...0e75d89.html" title="Last updated on Oct 26, 2022, 1:25 PM UTC (0e75d89)">Diff</a>


index.html
```diff
@@ -5362,8 +5362,8 @@ <h3>HTML Attribute State and Property Mappings</h3>
               </td>
               <td class="aria">
                 <div class="general">
-                  Either the <a data-cite="accname-1.2/#dfn-accessible-name">accessible name</a>, or the
-                  <a data-cite="accname-1.2/#dfn-accessible-description">accessible description</a>, or Not mapped (see Comments).
+                  Either contributes to the <a data-cite="accname-1.2/#dfn-accessible-name">accessible name</a>, the
+                  <a data-cite="accname-1.2/#dfn-accessible-description">accessible description</a>, or is not mapped (see comments).
```
spectranaut, line 5406: I don't mean to nitpick, but these 4 changes belong in a separate editorial PR, don't you think?

index.html
```diff
@@ -5799,12 +5870,12 @@ <h2>Accessible Name and Description Computation</h2>
     <p>The terms <a data-cite="accname-1.2/#dfn-accessible-name">accessible name</a> and <a data-cite="accname-1.2/#dfn-accessible-description">accessible description</a> are properties provided in all <a class="termref">accessibility APIs</a>. The name of the properties may differ across APIs but they serve the same function: as a container for a short (name) or longer (description) string of text. </p>
     <p>The <a href="#mapping_additional_nd_te" class="accname">text alternative computation</a> is used to generate both the <a data-cite="accname-1.2/#dfn-accessible-name">accessible name</a> and <a data-cite="accname-1.2/#dfn-accessible-description">accessible description</a>. There are different rules provided for several different types of <a class="termref" data-lt="element">elements</a>, <a class="termref" data-lt="node">nodes</a>, and combinations of markup.</p>
     <section>
-      <h3>`input type="text"`, `input type="password"`,` input type="search"`,` input type="tel"`, `input type="url"` and `textarea` Element</h3>
+      <h3>`input type="text"`, `input type="password"`, `input type="search"`, `input type="tel"`, `input type="url"` and `textarea` Element</h3>
```
spectranaut: can these and the updates from 1.1->1.2 also be in a separate editorial PR, and not be held back by any discussion of value? :)

index.html
```diff
@@ -5644,31 +5644,118 @@ <h3>HTML Attribute State and Property Mappings</h3>
...
+                <a data-cite="html/input.html#url-state-(type=url)">`type=url`</a>;
+                <a data-cite="html/input.html#week-state-(type=week)">`type=week`</a>;
+              </td>
+              <td class="aria">Not mapped</td>
+              <td class="ia2">See comments</td>
+              <td class="uia">See comments</td>
+              <td class="atk">See comments</td>
+              <td class="ax">See comments</td>
+              <td class="comments">
+                If specified, expose the value of the attribute as the initial value entry of the `input` element.
```
spectranaut, line 5767: Specifically, what is the "value entry of the `input` element"?

Do we have the following specified somewhere: if, in one of these inputs, a user has entered text, how that text surfaced in the different accessibility APIs? If so we should link to it if not we should add it?

index.html
```diff
@@ -5644,31 +5644,118 @@ <h3>HTML Attribute State and Property Mappings</h3>
...
+                <a class="core-mapping" href="#ariaValueNow">`aria-valuenow`</a> &amp;
+                <a class="core-mapping" href="#ariaValueText">`aria-valuetext`</a>
+              </td>
+              <td class="ia2">Use WAI-ARIA mapping</td>
+              <td class="uia">Use WAI-ARIA mapping</td>
+              <td class="atk">Use WAI-ARIA mapping</td>
+              <td class="ax">Use WAI-ARIA mapping</td>
+              <td class="comments">
+                <p>
+                  If specified, expose the value of the attribute as the initial value of the `input`. User Agents MUST expose the modified value of the input to the <a class="termref">accessibility tree</a>. User Agents MAY expose the initial or modified value as a human readable text alternative. E.g., `#ff0000` as a localized text alternative of "Red".
```
spectranaut, line 5738: Do these user agents MUSTs apply to the other inputs as well?

spectranaut (commented):
Still trying to catch up and get all the context for this change, i have a few questions to start!

index.html
spectranaut: As we discussed, let's use the same language here as in the HTML spec: https://html.spec.whatwg.org/multipage/input.html#attr-input-value

which I think would be (but check my work):
```
If specified, expose the content value attribute as the default value entry of the `input` element.
```

index.html
scottaohara: we talked about this.  bad habits. oops.  next time won't do!

index.html
spectranaut: After discussing, I think we agreed that "If specified, expose the value of the attribute as the initial value of the `input`. User Agents MUST expose the modified value of the input to the <a class="termref">accessibility tree</a>." belongs in it's own section about all inputs that have attributes (like `value` or `checked`) who's default value is set by the `content attribute`. See the spec: https://html.spec.whatwg.org/multipage/input.html#attr-input-value

This can happen here or in a follow up.

index.html
cookiecrook, line 5406: Editorial: I think this is phrased ambiguously:

Intention: "either it's:
- mapped (to name or description) or 
- not mapped."

As phrased with the 3-part serial comma: "either it's mapped to:
- name,
- description, or 
- not mapped" (seems like ~"mapped to not mapped")

index.html
spectranaut: Can we not say what these mappings are though? Do we know how the accessibility API for the different platforms surface the value of a `input [type="email"]` etc? Like AXValue for Mac?

spectranaut (changes requested):
From our conversation!

index.html
cookiecrook: Substantive: I think there are scenarios where implementations may still wish to map it even if it's not used by name or description. Leaving that follow-on mapping optional before the final "is not mapped."

For example, I think once aria-description is implemented, browsers would still want to expose the title here as AXHelp in AX API.
```html
<button aria-label="foo" aria-description="bar" title="bop">bif</button>
```

cookiecrook (changes requested):

rahimabdi:
@scottaohara I tried to migrate this one but it is indeed complex; it looks like there are several changes that have already been made in spec (e.g., specifying reset, button for <input> type) so parts of this PR may no longer be needed. The spec formatting has also changed since 2022 so I was unsure if you wanted to keep what we currently have in spec (e.g., input type=reset, type=submit as separate items) vs. this original PR where all of these would be contained in a single table.

If it's OK, I'll leave this one to you to determine what actually needs to be migrated to the mono repo.

scottaohara:
migrated to https://github.com/w3c/aria/pull/2373.  closing this pr

PR #402, opened 2022-05-04
clarify output element naming

### LINKED ISSUE #225: output element is a labelable element
[The `output` element](https://html.spec.whatwg.org/multipage/form-elements.html#the-output-element) is a labelable element and as such, its naming algorithm should be updated to indicate this.

Presently the algorithm is:
1. `aria-label` and `aria-labelledby`
2. subtree
3. `title`
4. no name

Should be updated to:
1. `aria-label` and `aria-labelledby`
2. name from `label` element
3. `title`
4. no name

edit: subtree shouldn't be considered for name.    name AND subtree need to be announced by the `output` live region and should not just announce name.

scottaohara:
closes #225

- Adds that the `label` element can name the output element
- Removes the step saying that `output` can get its name from its subtree.  This was incorrect, the name and content of the `output` need to be treated differently, as the name should not overwrite the subtree.  This aligns the element with the `status` role definition in the ARIA spec, where name is provided by author alone.
- removes the `localized control type: output` from the `output` role mapping table.  Should just be `status`, as the implicit role already specifies.

**Question**: is it worth adding a note or some normative text to call out that the name (if provided) and subtree are expected to (both) be announced as part of a live region event?

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/html-aam/pull/402.html" title="Last updated on Mar 8, 2023, 4:24 PM UTC (4b0c53b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/html-aam/402/a3cf752...4b0c53b.html" title="Last updated on Mar 8, 2023, 4:24 PM UTC (4b0c53b)">Diff</a>


index.html
```diff
@@ -2406,13 +2406,23 @@ <h3>HTML Element Role Mappings</h3>
...
+              <td class="uia"><div class="general">Use WAI-ARIA mapping</div></td>
+              <td class="atk">
+                <div class="general">Use WAI-ARIA mapping</div>
+                <div class="relations">
+                  <span class="type">Relations:</span> `ATK_RELATION_LABELLED_BY` with associated `label` element
+                </div>
+              </td>
+              <td class="ax">
+                <div class="general">Use WAI-ARIA mapping</div>
+                <td class="ax">`AXDescription: value from associated `label` element subtree.</td>
```
cookiecrook: For future reference, pretty much all instances to `AXTitle` or `AXDescription` should be replaced with `AXLabel`. There was a legacy distinction in the original Mac OS X Accessibility Protocol that was later simplified. 
```suggestion
                <td class="ax">`AXLabel: value from associated `label` element subtree.</td>
```

cookiecrook (approved):
Approve with attached change suggestion.

index.html
cookiecrook: It also may be worth considering whether to list these using a more common developer reference, like `accessibilityLabel`. If @aleventhal or other non-Apple implementors have a preference, please advise. Thanks.
https://developer.apple.com/documentation/objectivec/nsobject/1615181-accessibilitylabel

scottaohara: Unless you object, @cookiecrook, I'm going to leave this as is since there's a larger find and replace issue I filed to fix this more broadly (#450).

That way, if this is instead changed to `accessibilityLabel`, then this instance doesn't need to be changed again.

spectranaut: sort of related, you should fix the missing backtick for AXDescription

spectranaut (approved):
This looks good to me, provided you add that backtick I mentioned.

about your question: "is it worth adding a note or some normative text to call out that the name (if provided) and subtree are expected to (both) be announced as part of a live region event?"

It seems to me like that should be in the ARIA spec, right? I just did a quick look and couldn't see anything

index.html
```diff
@@ -2404,13 +2404,23 @@ <h3>HTML Element Role Mappings</h3>
...
```
```suggestion
                <td class="ax">`AXDescription`: value from associated `label` element subtree.</td>
```

PR #410, opened 2022-05-31
Addition: inert attribute mappings

### LINKED ISSUE #295: Inert subtrees mapping
Need to spec out how inert subtrees are exposed to AT. Dialog-backed inert content perhaps makes no interest for screen readers and may be safely cut off, however screen magnifiers might need it. HTML:inert, which is re-added in [1], seems adds a bunch of new use cases which vary from offscreen accessible elements to image carousel controls, thus I'm not quite sure what mapping could make the best experience for AT.

Note, inert subtrees spec [2] allows user agents to override the restrictions on search and text selection in inert subtrees, which may speak out against cut-off approach.

Perhaps, the safest thing to do is to expose an accessible state or object attribute in case of ATK/IA2 on inert subtrees to let AT to decide how to deal with inert content.

Thoughts?

cc'ing @jcsteh, @MarcoZehe, @alice, @cookiecrook

[1] https://github.com/alice/html/tree/re-add-inert
[2] https://html.spec.whatwg.org/multipage/interaction.html#inert-subtrees

### LINKED ISSUE #399: add inert attribute
https://html.spec.whatwg.org/multipage/interaction.html#the-inert-attribute

related to #295

scottaohara:
closes #295

add `inert` attribute and mappings.

wondering if these need to call out that subtree elements should not be focusable re: their mappings, or if the comment here is sufficient?  any other input on additional mappings that may need to be called out here are welcome.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/html-aam/pull/410.html" title="Last updated on Mar 28, 2023, 5:04 PM UTC (f362e5e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/html-aam/410/e8d3b12...f362e5e.html" title="Last updated on Mar 28, 2023, 5:04 PM UTC (f362e5e)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -4292,8 +4292,8 @@
               <td class="ax">Use WAI-ARIA mapping</td>
               <td class="comments">
                 <p>An inert element or the elements of an inert element's subtree cannot receive keyboard focus, nor will they respond to pointer events.</p>
-                <p>A `dialog` element can escape inertness of its inert ancestor when the `dialog` is modal and rendered in the browser's 
-                  <a href="https://fullscreen.spec.whatwg.org/#top-layer">top layer</a>.</p>
+                <p>A `dialog` element can escape an inert subtree when the `dialog` is in the modal state, and rendered in the browser's 
+                  <a href="https://fullscreen.spec.whatwg.org/#top-layer">top layer</a>.  A modal dialog, or its <a>flat tree</a> descendants can also become <a>inert</a>.  See <a data-cite="interaction.html#modal-dialogs-and-inert-subtrees">Modal dialogs and inert subtrees</a></p>
               </td>
             </tr>
             <tr tabindex="-1" id="att-indeterminate">
--- a/index.html
+++ b/index.html
@@ -4251,18 +4251,19 @@
               <td class="elements">
                 <a data-cite="html/interaction.html#the-inert-attribute">HTML elements</a>
               </td>
-              <td class="aria"><a class="core-mapping" href="#ariaHiddenTrue">`aria-hidden="true"`</a></td>
-              <td class="ia2">Use WAI-ARIA mapping</td>
-              <td class="uia">Use WAI-ARIA mapping</td>
-              <td class="atk">Use WAI-ARIA mapping</td>
-              <td class="ax">Use WAI-ARIA mapping</td>
+              <td class="aria">Not Mapped</td>
+              <td class="ia2">See comments</td>
+              <td class="uia">See comments</td>
+              <td class="atk">See comments</td>
+              <td class="ax">See comments</td>
               <td class="comments">
-                <p>An inert element, nor the nodes if its subtree can receive keyboard focus, respond to pointer events, 
-                   or be otherwise reachable by assistive technology.
+                <p>
+                  Nodes that are inert are not exposed to an accessibility API.
                 </p>
-                <p>A `dialog` element can escape an inert subtree when the `dialog` is in the modal state, and rendered in the browser's 
-                  <a href="https://fullscreen.spec.whatwg.org/#top-layer">top layer</a> by use of the `showModal()` method. 
-                  See also <a data-cite="html/interaction.html#modal-dialogs-and-inert-subtrees">Modal dialogs and inert subtrees</a>.
+                <p class="note">
+                  Note: an inert node can have descendants that are not inert. For example, a
+                  <a data-cite="html/interaction.html#modal-dialogs-and-inert-subtrees">modal dialog</a>
+                  can escape an inert subtree.
                 </p>
               </td>
             </tr>
--- a/index.html
+++ b/index.html
@@ -4258,7 +4258,7 @@
               <td class="ax">See comments</td>
               <td class="comments">
                 <p>
-                  Nodes that are inert are not exposed to an accessibility API.
+                  Nodes that are <a data-cite="html/interaction.html#inert">inert</a> are not exposed to an accessibility API.
                 </p>
                 <p class="note">
                   Note: an inert node can have descendants that are not inert. For example, a
--- a/index.html
+++ b/index.html
@@ -6189,6 +6189,7 @@
       <section>
         <h4>Substantive changes since moving to the <a href="https://www.w3.org/WAI/ARIA/">Accessible Rich Internet Applications Working Group</a> (03-Nov-2019)</h4>
         <ul>
+          <li>23-Mar-2023: Add `inert` attribute mapping. See <a href="https://github.com/w3c/html-aam/pull/410">GitHub PR 410</a>.</li>
           <li>08-Mar-2023: Update `hgroup` element to be mapped to `role=group`. See  <a href="https://github.com/w3c/html-aam/pull/398">GitHub PR 398</a>.</li>
           <li>08-Mar-2023: Clarify naming algorithm for `output` element. See <a href="https://github.com/w3c/html-aam/pull/402">GitHub PR 402</a>.</li>
           <li>12-Dec-2022: Revise mapping for `s` element to be `role=deletion`. See <a href="https://github.com/w3c/html-aam/pull/442">GitHub PR 442</a>.</li>
--- a/index.html
+++ b/index.html
@@ -4280,6 +4280,22 @@
               <td class="ax"><div class="general">Not mapped</div></td>
               <td class="comments"></td>
             </tr>
+            <tr tabindex="-1" id="att-inert">
+              <th>`inert`</th>
+              <td class="elements">
+                <a data-cite="html/interaction.html#the-inert-attribute">HTML elements</a>
+              </td>
+              <td class="aria"><a class="core-mapping" href="#ariaHiddenTrue">`aria-hidden="true"`</a></td>
+              <td class="ia2">Use WAI-ARIA mapping</td>
+              <td class="uia">Use WAI-ARIA mapping</td>
+              <td class="atk">Use WAI-ARIA mapping</td>
+              <td class="ax">Use WAI-ARIA mapping</td>
+              <td class="comments">
+                <p>An inert element or the elements of an inert element's subtree cannot receive keyboard focus, nor will they respond to pointer events.</p>
+                <p>A `dialog` element can escape inertness of its inert ancestor when the `dialog` is modal and rendered in the browser's 
+                  <a href="https://fullscreen.spec.whatwg.org/#top-layer">top layer</a>.</p>
+              </td>
+            </tr>
             <tr tabindex="-1" id="att-indeterminate">
               <th>`indeterminate [IDL]`</th>
               <td class="elements">
--- a/index.html
+++ b/index.html
@@ -4291,9 +4291,13 @@
               <td class="atk">Use WAI-ARIA mapping</td>
               <td class="ax">Use WAI-ARIA mapping</td>
               <td class="comments">
-                <p>An inert element or the elements of an inert element's subtree cannot receive keyboard focus, nor will they respond to pointer events.</p>
+                <p>An inert element, nor the nodes if its subtree can receive keyboard focus, respond to pointer events, 
+                   or be otherwise reachable by assistive technology.
+                </p>
                 <p>A `dialog` element can escape an inert subtree when the `dialog` is in the modal state, and rendered in the browser's 
-                  <a href="https://fullscreen.spec.whatwg.org/#top-layer">top layer</a>.  A modal dialog, or its <a>flat tree</a> descendants can also become <a>inert</a>.  See <a data-cite="interaction.html#modal-dialogs-and-inert-subtrees">Modal dialogs and inert subtrees</a></p>
+                  <a href="https://fullscreen.spec.whatwg.org/#top-layer">top layer</a> by use of the `showModal()` method. 
+                  See also <a data-cite="html/interaction.html#modal-dialogs-and-inert-subtrees">Modal dialogs and inert subtrees</a>.
+                </p>
               </td>
             </tr>
             <tr tabindex="-1" id="att-indeterminate">


scottaohara:
Note: I've been drafting updates to the HTML spec for the inert subtree/attribute section.   I will reference that here when the PR is made.

index.html
```diff
@@ -4280,6 +4280,22 @@ <h3>HTML Attribute State and Property Mappings</h3>
...
+              <td class="elements">
+                <a data-cite="html/interaction.html#the-inert-attribute">HTML elements</a>
+              </td>
+              <td class="aria"><a class="core-mapping" href="#ariaHiddenTrue">`aria-hidden="true"`</a></td>
+              <td class="ia2">Use WAI-ARIA mapping</td>
+              <td class="uia">Use WAI-ARIA mapping</td>
+              <td class="atk">Use WAI-ARIA mapping</td>
+              <td class="ax">Use WAI-ARIA mapping</td>
+              <td class="comments">
+                <p>An inert element or the elements of an inert element's subtree cannot receive keyboard focus, nor will they respond to pointer events.</p>
```
spectranaut: Once you land guidance in the HTML spec, do you need to reiterate this here, rather than just link? Same with your question in the description: `wondering if these need to call out that subtree elements should not be focusable re: their mappings, or if the comment here is sufficient?`

Or does having a focusable subtree element have an effect on the mapping? In which case you should call it out. I think the answer is no, just checking.

index.html
```diff
@@ -4280,6 +4280,22 @@ <h3>HTML Attribute State and Property Mappings</h3>
...
+                <p>A `dialog` element can escape inertness of its inert ancestor when the `dialog` is modal and rendered in the browser's 
+                  <a href="https://fullscreen.spec.whatwg.org/#top-layer">top layer</a>.</p>
```
spectranaut: What does "escape inertness" mean for html-aam implementations? That the attribute should be ignored? Maybe we should say that?

spectranaut (changes requested):
Couple questions :)

index.html
scottaohara: the guidance to add is mainly author guidance, not implementor.  and the implementor guidance there that i am going to touch on is reflective of the mapping here.  further author guidance will be written for MDN as well... so any further guidance here really should just be implementor focused.

index.html
```diff
@@ -4280,6 +4280,22 @@ <h3>HTML Attribute State and Property Mappings</h3>
...
+                <a data-cite="html/interaction.html#the-inert-attribute">HTML elements</a>
+              </td>
+              <td class="aria"><a class="core-mapping" href="#ariaHiddenTrue">`aria-hidden="true"`</a></td>
+              <td class="ia2">Use WAI-ARIA mapping</td>
+              <td class="uia">Use WAI-ARIA mapping</td>
+              <td class="atk">Use WAI-ARIA mapping</td>
+              <td class="ax">Use WAI-ARIA mapping</td>
+              <td class="comments">
+                <p>An inert element or the elements of an inert element's subtree cannot receive keyboard focus, nor will they respond to pointer events.</p>
+                <p>A `dialog` element can escape inertness of its inert ancestor when the `dialog` is modal and rendered in the browser's
```
```suggestion
                <p>A `dialog` element can escape an inert subtree when the `dialog` is in the modal state, and rendered in the browser's 
```

index.html
```suggestion
                  <a href="https://fullscreen.spec.whatwg.org/#top-layer">top layer</a>.  A modal dialog, or its <a>flat tree</a> descendants can also become <a>inert</a>.  See <a data-cite="interaction.html#modal-dialogs-and-inert-subtrees">Modal dialogs and inert subtrees</a></p>
```

scottaohara: @spectranaut was using language from HTML.  have revised and expanded on that part of the comment, linking back to HTML for more info.

index.html
scottaohara: re: my question about focusing.  I personally do not think that needs to be called out here.  it's already mentioned in the HTML spec, so removing the ability for the element/its subtree to be focused in effect removes the mappings related to being focusable.  but if i'm wrong on that, then i'd want to be told what needs to be mentioned here beyond the comment reiterating that the element and/or would-be focusable elements in its subtree cannot receive focus.

scottaohara:
@spectranaut thank you for the review

index.html
spectranaut: Is it odd to add author guidance to the HTML-AAM? I don't remember seeing a lot of author guidance in any of the AAMs, seems like that would mostly belongs in ARIA or HTML? you are the editor, though, so I defer to you :)

index.html
spectranaut: I'm afraid I'm still confused, if I had to implement this I wouldn't be positive what to do. "aria-hidden" prunes the subtree, so does this mean we have to go searching through the subtree of all aria-hidden elements for a dialog element with aria-modal="true", and check whether it is in the "top layer"?

Also what are the "flat tree" descendants of something?

index.html
scottaohara: by adding author guidance i meant that *that* guidance is going into the HTML spec and even further guidance would go to MDN.  No author guidance would go here.

additionally, maybe my previous comment caused confusion?  i merely meant that i was going to reference the PR for HTML's author guidance here in this PR thread... not in the update to HTML AAM.  Sorry if that was unclear/caused this thread of discussion.

index.html
scottaohara: those were supposed to be links to their definitions that i hadn't gotten around to getting yet.

but, per your feedback, i've revised/reduced the commentary here again.  This is just for the mapping that needs to occur for the attribute, which is aria-hidden=true.  The rest of the comments are meant to signify expected behaviors.  If more clarity is needed for some of the concepts/terms i was referencing that were defined in other specs, then we should probably make them more clear there, rather than expanding on their meaning in comments here.

thank you for this, this was helpful to keep me from over complicating this.

scottaohara:
cc @cookiecrook and @jcsteh for feedback and anything else you think needs to be called out in this spec for `inert`?  Please see the companion HTML PR for `inert` that incorporates your feedback into that specification.
https://github.com/whatwg/html/pull/8122

zcorpan:
Hmm, so `aria-hidden=true` make the whole subtree be hidden, even if a descendant has `aria-hidden=false` (like CSS `display: none`). Maybe we need something in ARIA that is akin to CSS `visibility`, which can be `hidden` on an element and a descendant can still be shown with `visibility: visible`, to property define a mapping for `inert`?

Edit: or at least a spec hook to do so, which html-aam can use to explain modal `dialog` in an inert subtree.

zcorpan:
Demo of modal dialog inside `<div inert>`:
http://software.hixie.ch/utilities/js/live-dom-viewer/saved/11311

In Chrome, Safari, Firefox, the button "show (in inert div)" is not clickable, but the "close" button in the modal dialog is.

JAWS-test:
> In Chrome, Safari, Firefox, the button "show (in inert div)" is not clickable, but the "close" button in the modal dialog is.

On Windows I get a different result: in Firefox both buttons are clickable and the inert area is fully passed to the Accessbility API

zcorpan:
Sorry, I meant Firefox Nightly. (`inert` is not yet enabled on release because this spec change is not yet implemented, see https://bugzilla.mozilla.org/show_bug.cgi?id=1767561 )

What I mean to show with the demo is that a modal `dialog` should itself be not-inert, even if it has an ancestor with an `inert` attribute.

index.html
```diff
@@ -4250,6 +4250,26 @@ <h3>HTML Attribute State and Property Mappings</h3>
...
+              <td class="uia">Use WAI-ARIA mapping</td>
+              <td class="atk">Use WAI-ARIA mapping</td>
+              <td class="ax">Use WAI-ARIA mapping</td>
+              <td class="comments">
+                <p>An inert element, nor the nodes if its subtree can receive keyboard focus, respond to pointer events, 
+                   or be otherwise reachable by assistive technology.
+                </p>
+                <p>A `dialog` element can escape an inert subtree when the `dialog` is in the modal state, and rendered in the browser's 
+                  <a href="https://fullscreen.spec.whatwg.org/#top-layer">top layer</a> by use of the `showModal()` method. 
+                  See also <a data-cite="html/interaction.html#modal-dialogs-and-inert-subtrees">Modal dialogs and inert subtrees</a>.
```
cookiecrook: How should a `<div role="dialog">` or a `<div role="menu">` achieve the same?

scottaohara: possibly by using the popover attribute, though i don't believe that functionality is available right now.  we should open an issue for that separate from this.

cookiecrook: As I read it, popover is not intended for modals like `<div role="dialog" aria-modal="true">`. The Dialog API is the only option for a modal. Regardless, I agree that's addressable as a separate issue.

cookiecrook (approved):

index.html
zcorpan: Right, ARIA would need a way of unhiding a child, see https://github.com/w3c/html-aam/pull/410#issuecomment-1435772527

I can file a spec issue.

zcorpan: You had filed one already :) https://github.com/w3c/aria/issues/1256

cookiecrook: That's specifically for aria-hidden. I don't think aria should be able to un-inert an inert descendant.

zcorpan: I'm now confused what you mean with your first message in this thread.

scottaohara: dialog is the only element, when shown as a modal, which can escape its parent inert element and become re-exposed in the top layer.

so, per James's question, there is presently no way for someone to invoke an accessible ARIA dialog (modal or otherwise), or a menu if someone wanted to trap focus/screen readers in those elements, if it was contained within an inert container.

though popover is not allowed for HTML modal dialogs, the attribute does send an element to the top layer, so my mention of the attribute is that _it_ could be a way for someone to do something like:

```
<div inert> <!-- attribute added when nested button is activated -->
  <button aria-haspopup=menu ...>
     invokes menu and sets parent to inert
  </button>
  <div role=menu popover ...> 
     <!-- menu items here and even though within an inert container, 
             since its in the top layer now, it COULD now be accessible --> 
   </div>
</div>
```

somewhere in some github thread that i've since lost track of, when implementation of inert was still being discussed, I had mentioned it would be great if `inert=false` would un-inert a child node of an inert parent.  I don't understand the rational as to why that wasn't entertained further, other than potential implementation complexity.  But back to James's original question, because of that decision there is no way to do what James was asking with HTML now, and `aria-hidden=false` wouldn't/shouldn't be expected to be a way to achieve this, as even if it did re-expose a part of the inert content, by itself it wouldn't undo the browser preventing click/focus to go to the aria-hidden=false content (as applicable if it were to consist of actionable elements)

scottaohara: marking this thread as unresolved just to make sure this related discussion about inert/aria-hidden=false isn't missed by those being directed to this issue.

scottaohara:
@zcorpan per your previous comment
>Edit: or at least a spec hook to do so, which html-aam can use to explain modal dialog in an inert subtree.

with linked example and 
>What I mean to show with the demo is that a modal dialog should itself be not-inert, even if it has an ancestor with an inert attribute.

to my understanding that's already specified in HTML 6.3.1 Modal dialogs and inert subtrees - reference in this PR.  So i don't think we need anything else?

Re: your example contains a button and the dialog in the inert subtree, the dialog is spec'd to escape it. "subject escapes inertness of ancestors" so that's why the button, and its contents are accessible.

followed on by the note
>The [dialog](https://html.spec.whatwg.org/multipage/interactive-elements.html#the-dialog-element) element's [showModal()](https://html.spec.whatwg.org/multipage/interactive-elements.html#dom-dialog-showmodal) method causes this mechanism to trigger, by [adding](https://fullscreen.spec.whatwg.org/#top-layer-add) the [dialog](https://html.spec.whatwg.org/multipage/interactive-elements.html#the-dialog-element) element to its [node document](https://dom.spec.whatwg.org/#concept-node-document)'s [top layer](https://fullscreen.spec.whatwg.org/#top-layer).

again, per my response in the comment thread with James, HTML could leverage this to allow popovers to also escape inert, since they too go to the top layer.  But per your example and the button that invokes such elements being within an inert parent - this would likely require the topic of `inert=false` to be re-opened.

Edit: however i also realize i may be reading way too much into the note and what "this mechanism" really means.  but if i'm wrong, maybe that could use some clarification?

zcorpan:
I think the HTML spec is correct about which nodes are inert and which are not, but this PR mapping the `inert` attribute to `aria-hidden=true` (which makes the element itself and all of its descendants hidden in the a11y tree) doesn't match the HTML spec because it means a descendant modal dialog (which should be uninerted) is still hidden in the a11y tree.

More correct would be to say something like "Nodes that are [inert](https://html.spec.whatwg.org/multipage/interaction.html#inert) are not exposed to an accessibility API. Note: an inert node can have descendants that are not inert (e.g. a modal `dialog` element in HTML). Note: This is similar to `aria-hidden`, except `aria-hidden=true` will unconditionally hide all descendants."

scottaohara:
@zcorpan re:
>Note: an inert node can have descendants that are not inert (e.g. a modal dialog element in HTML).

will that be true for any other element aside from modal dialogs?  my understanding is that's not true right now, and per my mention of `inert=false` not being entertained, is this actually something that would happen?

Seems the PR could either be updated to specifically mention that if an element (a modal dialog) escapes being inert, then it should no longer be considered aria-hidden=true.  Or, the PR is further adjusted to be more inline with your suggestion that inert doesn't directly map to aria-hidden=true, but it and its subtree behave similarly, but descendants of an inert parent can be marked as not inert - e.g., a modal dialog.

@cookiecrook @aleventhal thoughts on the above?  and specifically, are there any other platform mappings we'd need to note here if the attribute isn't just referring to aria-hidden's mappings in core aam?

cookiecrook:
I think Simon's normative suggestion and first note are correct:

> Nodes that are [inert](https://html.spec.whatwg.org/multipage/interaction.html#inert) are not exposed to an accessibility API. Note: an inert node can have descendants that are not inert (e.g. a modal dialog element in HTML).

But I would not include the second note:

> ~~Note: This is similar to aria-hidden, except aria-hidden=true will unconditionally hide all descendants.~~

As that may change further with https://github.com/w3c/aria/issues/1256 and the implementations don't currently align on that statement either.

zcorpan:
@scottaohara it might be true in the future if we introduce other ways to escape inertness.

scottaohara:
@cookiecrook @zcorpan thank you both.  i'll work on getting this updated.

should we actually update the ARIA spec to mention inert in https://www.w3.org/TR/wai-aria-1.2/#tree_exclusion

that seems it might be a good place for a longer explanation of how to handle inert / potential children which are no longer inert, which could then be linked to from html aam / other aams which might need it.

index.html
```diff
@@ -4246,6 +4246,27 @@ <h3>HTML Attribute State and Property Mappings</h3>
...
+                <a data-cite="html/interaction.html#the-inert-attribute">HTML elements</a>
+              </td>
+              <td class="aria">Not Mapped</td>
+              <td class="ia2">See comments</td>
+              <td class="uia">See comments</td>
+              <td class="atk">See comments</td>
+              <td class="ax">See comments</td>
+              <td class="comments">
+                <p>
+                  Nodes that are inert are not exposed to an accessibility API.
```
```suggestion
                  Nodes that are <a data-cite="html/interaction.html#inert">inert</a> are not exposed to an accessibility API.
```

zcorpan (approved):

cookiecrook (approved):

index.html
alice: > it would be great if `inert=false` would un-inert a child node of an inert parent

Coming in very late, and obviously this has no bearing on the outcome of this PR, but I talked a bit about the rationale here: https://www.igalia.com/chats/alice-and-rob (you can search for "one way" in the transcript rather than listen to the whole rambling conversation).

scottaohara: @alice thank you for the link and background!

PR #423, opened 2022-07-08
Correction: add missing label elm to 'button' naming steps

### LINKED ISSUE #357: Button accessibility name doesn't indicate label element can be used
The label element seems to be usable to label buttons but this doesn't seem to be list in the accessible name calculation.

scottaohara:
adds label element to `button`, `input type=button|submit|reset|image`.

This is a long overdue update and there could be more clarity here.  However, getting this perfect has long stood in the way of getting this done, and I would rather get this gap resolved to match reality, and then better describe this in a follow revision to these naming steps.

Closes #357 - companion PR to #402

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/html-aam/pull/423.html" title="Last updated on Jul 19, 2022, 2:00 PM UTC (b7aa00b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/html-aam/423/4310e8d...b7aa00b.html" title="Last updated on Jul 19, 2022, 2:00 PM UTC (b7aa00b)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -5927,6 +5927,9 @@
           <li>
             If the control has an <a data-cite="wai-aria-1.2/#aria-label">`aria-label`</a> or an <a data-cite="wai-aria-1.2/#aria-labelledby">`aria-labelledby`</a> attribute the <a data-cite="accname-1.2/#dfn-accessible-name">accessible name</a> is to be calculated using the algorithm defined in <a href="" class="accname">Accessible Name and Description: Computation and API Mappings</a>.
           </li>
+          <li>
+            Otherwise use the associated `label` element(s) <a data-cite="accname-1.2/#dfn-accessible-name">accessible name</a>(s) - if more than one `label` is associated; concatenate by DOM order, delimited by spaces.
+          </li>
           <li>Otherwise use the `value` attribute.</li>
           <li>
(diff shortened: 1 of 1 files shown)


cookiecrook (commented):
I feel like this HTML-AAM specific algorithm may start to get to complicated unless it's explicit where each of these steps fits into the main AccName algorithm. For example:

- Step 1 (aria-label, etc here) is redundant with AccName...
- Steps 2 and 3 (label el, value attr) fits into the "host language label mechanism" portion of AccName
- Steps 4 and 5 (title and placeholder attrs) fit into the host language tooltip mechanism portion of AccName.
- then the final step would be the element-specific fallback label

scottaohara:
@cookiecrook yes, and this is where some of the cleanup i'm trying to do in another future PR would come into play.

But there are some elements that don't fit neatly with the way that accName is setup.  E.g., there are some elements where name from content (accName F) takes precedent over the `title` attribute... but accName lists label element and title attribute under accName step D.  That's not right in all cases, nor should it be.

cookiecrook:
Okay then. I approve this diff as-is. Thanks for the follow-up.

cookiecrook (approved):

PR #430, opened 2022-07-13
Adds language to expand 'not mapped' definition

### LINKED ISSUE #151: Need language/mapping for "not-mapped" elements which must be exposed due to attributes

scottaohara:
closes #151

While there are many elements that would have originally fallen under this issue... on a long enough timeline everything maps to `generic`.

So is the resolution to this issue.  There are a few remaining HTML elements that are not mapped because they are not rendered by browsers.  For the few instances of these where users can force render them (e.g., making a `style` element `display: block` for reasons) then these elements would and should be largely exposed as `generic` elements.  From quick review, the only one in question here of not mapping to `generic` is `link`. But also, if you're force rendering a `link` element... then...

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/html-aam/pull/430.html" title="Last updated on Oct 26, 2022, 1:04 PM UTC (fc96fd2)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/html-aam/430/36fa626...fc96fd2.html" title="Last updated on Oct 26, 2022, 1:04 PM UTC (fc96fd2)">Diff</a>


index.html
```diff
@@ -217,8 +217,8 @@ <h3>Exposing HTML Features That Do Not Directly Map to Accessibility APIs</h3>
...
-        <li>"Not mapped" (Not Applicable) means the element does not need to be exposed via an <a class="termref">accessibility API</a>. This is usually because the element is not displayed as part of the user interface.</li>
+        <li>HTML elements with implicit WAI-ARIA role semantics MUST be mapped to platform <a class="termref">accessibility APIs</a> according to the identified WAI-ARIA role mapping as defined in the [[core-aam-1.2]] specification.</li>
+        <li>"Not mapped" means the element does not need to be exposed via an <a class="termref">accessibility API</a>. This is usually because the element is not displayed as part of the user interface. However, authors can force some of these elements to be rendered. For instance, by overriding user agent styles to render elements that would have been otherwise set to `display: none`. In these cases, the implicit mapping of such elements SHOULD be <a class="core-mapping" href="#role-map-generic">`generic`</a>.</li>
```
cookiecrook:
```suggestion
        <li>"Not mapped" means the element does not need to be exposed via an <a class="termref">accessibility API</a>. This is usually because the element is not displayed as part of the user interface. However, authors can force some of these elements to be rendered. For instance, by overriding user agent styles to render elements that would have been otherwise set to `display: none`. In these cases, the user agent SHOULD map such elements to the role of <a class="core-mapping" href="#role-map-generic">`generic`</a>.</li>
```

cookiecrook: Can't have a RFC-2119 keyword in passive voice. Define who the requirement is for.

cookiecrook (approved):

cookiecrook (approved):

index.html
scottaohara: thank you

Issue #278, opened 2020-02-03
Description Computation for buttons not consistent


JAWS-test:
I wonder why the Description Computation is not consistent for button and I can't find any reason for the differences in [ACC Name](https://w3c.github.io/accname/):
* [Button with `input`](https://w3c.github.io/html-aam/#input-type-button-input-type-submit-and-input-type-reset-accessible-description-computation): visual text content can be description (`value`)
* [Button with `summary`](https://w3c.github.io/html-aam/#summary-element-accessible-description-computation): Text content can be description
* [Button with `button`](https://w3c.github.io/html-aam/#button-element-accessible-description-computation): Text content cannot be description

scottaohara:
With Chrome, if you review the accessibility tree of the elements in [this test case](https://scottaohara.github.io/testing/buttons/desc.html), the `value` of an `input type=button` with an `aria-label` is exposed as a description.  As is the visible text of a `summary` with an `aria-label`.

In Webkit the visible text of a `summary` is exposed to the a11y tree (noted as `title` content in Apple's Accessibility Inspector) and can be read by VO if one were to navigate into the `summary` element (`summary` is not exposed as a `button` in Webkit).

Firefox doesn't expose the `value` of an `input type=button` or subtree of a `summary` as description content.

In summary: 
* Chrome and Chromium Edge with JAWS or NVDA will announce the descriptions.
* macOS Safari + VO will allow access to the visible text/content of the `summary`, but doesn't use the `value` of a `input type=button` as description content.
* Firefox doesn't follow the steps of these description algorithms.

JAWS-test:
Regardless of the test results, I wonder if this should not be unified and if Accname should be updated

scottaohara:
Don't see a good way for unification since each of these elements have variations in how they're used.  e.g. `value` is not used for `summary`, and doesn't expose name/desc information for the `button` element.

JAWS-test:
There are two possibilities:
* For all buttons the visible label can become a description
* For no button the visible label is used for description

scottaohara:
the first possible simplification would result in
```html
<button aria-label=close>
  X
</button>
```
having a description of "X".  I highly doubt we'd want to do this.


Which leaves the second option, which does bring these buttons inline with how other elements work.  But I'm curious of any reasoning as to why these were outlined this way to begin with before anything is done here.

JAWS-test:
I would also prefer the second option

JAWS-test:
[...]

PR #433, opened 2022-08-29
accessible description computation revision

scottaohara:
This PR removes the various individual element accessible description computations and instead combines them into a single stepped computation listing, as apart from `summary`, `table` and `input type=button|reset|submit`, they were all the same.

The revised accDescription still calls out these special cases (so nothing in the computation has changed here).  We should probably discuss these again in the future (for instance, see https://github.com/w3c/html-aam/issues/278)

This PR also includes `aria-description` as being part of the first step in the accDescription computation (similar to how aria-label/aria-labelledby are mentioned as the first step in the element naming computations).  AccName defines the priority of these two attributes in the accDesc computation.

This PR is related to https://github.com/w3c/accname/pull/69

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/html-aam/pull/433.html" title="Last updated on Oct 26, 2022, 1:16 PM UTC (4f2a8b6)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/html-aam/433/96ab938...4f2a8b6.html" title="Last updated on Oct 26, 2022, 1:16 PM UTC (4f2a8b6)">Diff</a>


index.html
```diff
@@ -6372,20 +6097,29 @@ <h4>Text-level Element Accessible Name Computation</h4>
...
-          </li>
-        </ol>
-      </section>
+    </section>
+    <section id="accdesc-computation">
+      <h3>Accessible Description Computation</h3>
+      <p>An <a data-cite="accname-1.2/#dfn-accessible-description">accessible description</a> MAY be provided to any HTML element that is a valid child of the `body` element.</p>
+      <ol>
+        <li>
+          If the element has an <a data-cite="wai-aria-1.2/#aria-describedby">`aria-describedby`</a> or <a data-cite="wai-aria-1.2/#aria-description">`aria-description`</a> attribute the <a data-cite="accname-1.2/#dfn-accessible-description">accessible description</a> is to be calculated using the algorithm defined in <a href="" class="accname">Accessible Name and Description: Computation and API Mappings</a>.
```
scottaohara: aria-description is not in ARIA 1.2 - so this needs to change to https://w3c.github.io/aria/#aria-description

scottaohara: though, it seems that HTML AAM should probably just always link to the latest drafts of docs, rather than different versions of the specs

MelSumner (changes requested):
This PR certainly simplifies things and makes it a lot clearer. 

I think you made a valid point with

> it seems that HTML AAM should probably just always link to the latest drafts of docs, rather than different versions of the specs

Should the links be updated before approve/merge?

index.html
```diff
@@ -6372,20 +6097,29 @@ <h4>Text-level Element Accessible Name Computation</h4>
...
+        </li>
+        <li>
+          Otherwise, if the <a data-cite="accname-1.2/#dfn-accessible-description">accessible description</a> is still empty, then:
+          <ul>
+            <li>If a `table` element, has a <a href="https://dom.spec.whatwg.org/#concept-tree-child">child</a> that is a `caption` element and it was not used as the <a data-cite="accname-1.2/#dfn-accessible-name">accessible name</a>, then use the subtree of the first such element.</li>
```
```suggestion
            <li>
              If the element is a `table` element:
              <ol>
                <li>If the `table` element has a <a href="https://dom.spec.whatwg.org/#concept-tree-child">child</a> that is a `caption` element, then use the subtree of the first `caption` if the `caption` was not used as the <a data-cite="accname-1.2/#dfn-accessible-name">accessible name</a>.</li>
              </ol>
            <li>
```

index.html
```diff
@@ -6372,20 +6097,29 @@ <h4>Text-level Element Accessible Name Computation</h4>
...
+            <li>If a `summary` element, use the element's subtree if it was not used as the <a data-cite="accname-1.2/#dfn-accessible-name">accessible name</a>.</li>
```
```suggestion
            <li>
              If the element is a `details` element:
              <ol>
                <li>Use the element's `summary` child if the `summary` was not used as the <a data-cite="accname-1.2/#dfn-accessible-name">accessible name</a>.</li>
              </ol>
            </li>
```

index.html
```diff
@@ -6372,20 +6097,29 @@ <h4>Text-level Element Accessible Name Computation</h4>
...
+            <li>If an `input` element whose type attribute is the `button`, `submit` or `reset` state has a `value` attribute and it was not used as the <a data-cite="accname-1.2/#dfn-accessible-name">accessible name</a>, then use the content of the `value` attribute.</li>
```
```suggestion
           <li>
             If the element is an `input` element whose type attribute is the `button`, `submit` or `reset`:
             <ol>
               <li>Use the contents of the html `value` attribute if the `value` attribute was not used as the <a data-cite="accname-1.2/#dfn-accessible-name">accessible name</a>.</li>
             </ol>
           </li>
```

spectranaut (changes requested):
what do you think of these changes @scottaohara? when I first read you suggested change is stuck me as not as clear as the original table in the corrosponding accname pr.

index.html
spectranaut: I feel like this makes me feel like all of our specs should be evergreen? If you link from ARIA 1.3 to core-aam and then core-aam links back to the working draft? Seems odd to come back to a draft you didn't didn't start out in?

Anyway, for this, though, I agree.

stevefaulkner (approved):

scottaohara:
@spectranaut appreciate the suggestions.  i'm going to adjust just a couple of things, but yeh, that helps call out the element and its special condition better than as one lumpy sentence.

index.html
scottaohara: FYI, the details element does not get a name from the summary element. This was wrong in the core aam PR and why i changed it here.

spectranaut: ohhh I see thxx

index.html
```diff
@@ -6372,20 +6097,44 @@ <h4>Text-level Element Accessible Name Computation</h4>
...
+        <li>
+          If the element has an <a data-cite="wai-aria-1.2/#aria-describedby">`aria-describedby`</a> or <a data-cite="wai-aria-1.2/#aria-description">`aria-description`</a> attribute the <a data-cite="accname-1.2/#dfn-accessible-description">accessible description</a> is to be calculated using the algorithm defined in <a href="" class="accname">Accessible Name and Description: Computation and API Mappings</a>.
+        </li>
+        <li>
+          Otherwise, if the <a data-cite="accname-1.2/#dfn-accessible-description">accessible description</a> is still empty, then:
+          <ul>
+            <li>
+              If a `table` element:
+              <ul>
+                <li>has a <a href="https://dom.spec.whatwg.org/#concept-tree-child">child</a> `caption` element, use the subtree of the first `caption` element if it was not used as the <a data-cite="accname-1.2/#dfn-accessible-name">accessible name</a>.</li>
```
spectranaut: @scottaohara I think each bullet should be a full sentence thus the verbosity of my suggestions -- "If a table element" is not a full sentence and neither is "has a caption element, us the subtree of the first...."

but you are the editor so ultimately up to you :)

scottaohara: hmm... understand your comment. just seems that if more verbosity is needed to introduce each single item sub-list, then maybe they should just stay sentences instead of breaking them apart, so as not having to say "if the element is a table element" and then start the new sentence with "if the table element..." again.

spectranaut: I see -- I now that I'm looking I realize I don't even suggest full sentences, so maybe that isn't what I'm after. I think this comes down to a preference.

Personally, if I'm reading the spec, I'd want to read "If the element is a table element" rather than "if a table element". If I was reading this step, I'd panic a little at this line. My internal dialog: "If a table element.... if a table element... if WHAT a table element??????"  I want the spec to tell me WHAT specifically I'm checking to be a table element. Even if it is obvious from context clues.

For example in this change I suggested: https://github.com/w3c/html-aam/pull/433#discussion_r981593862

I want each bullet to be an actionable step with all the objects/parts of that step included in the language of the step. I'd prefer there to be repetition rather than anything left out.
- action: check if the element is a table element -> spec: "if the element is a `table` element"
- action: check if the table element has a caption and the caption has not been already used, use the subtree -> spec: "If the `table` element has a child that is a `caption` element, then use the subtree of the first `caption` if the `caption` was not used as the accessible name"

scottaohara: @spectranaut thanks again, and yeh, definitely seems a preference thing, so i'm trying to adjust a bit more to try and meet in the middle.

i've reverted back to sentences and instead adjusted the intro sentence to the sub-lists to try to make it more clear without adding in redundancies.

index.html
```diff
@@ -6372,20 +6097,35 @@ <h4>Text-level Element Accessible Name Computation</h4>
...
+          If the element has an <a data-cite="wai-aria-1.2/#aria-describedby">`aria-describedby`</a> or <a data-cite="wai-aria-1.2/#aria-description">`aria-description`</a> attribute the <a data-cite="accname-1.2/#dfn-accessible-description">accessible description</a> is to be calculated using the algorithm defined in <a href="" class="accname">Accessible Name and Description: Computation and API Mappings</a>.
+        </li>
+        <li>
+          Otherwise, if the <a data-cite="accname-1.2/#dfn-accessible-description">accessible description</a> is still empty, and the element is:
+          <ul>
+            <li>
+              a `table` element which has a <a href="https://dom.spec.whatwg.org/#concept-tree-child">child</a> `caption` element, use the subtree of the first `caption` element if it was not used as the <a data-cite="accname-1.2/#dfn-accessible-name">accessible name</a>.
+            </li>
+            <li>
+              `summary` element, use the element's subtree if it was not used as the <a data-cite="accname-1.2/#dfn-accessible-name">accessible name</a>.
```
```suggestion
              a `summary` element, use the element's subtree if it was not used as the <a data-cite="accname-1.2/#dfn-accessible-name">accessible name</a>.
```

Add an "a"

spectranaut (approved):
looks good to me now :)

scottaohara:
merging this.  This issue https://github.com/w3c/html-aam/issues/437 will be where the spec is updated to link to the latest editors' draft of the ARIA specs

PR #438, opened 2022-10-03
chore: add PR template

spectranaut:
Closes #????

 * [ ] Modified Web platform tests (link to pull request)

Implementation commitment:

 * [ ] WebKit (https://bugs.webkit.org/show_bug.cgi?id=)
 * [ ] Chromium (https://bugs.chromium.org/p/chromium/issues/detail?id=)
 * [ ] Gecko (https://bugzilla.mozilla.org/show_bug.cgi?id=)

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,11 @@
+<!--- IF EDITORIAL or CHORE, delete this template -->
+
+Describe Change Here!
+
+# Implementation
+
+* WPT tests: [LINK]()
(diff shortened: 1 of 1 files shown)


spectranaut:
oh wow I just noticed html already has a PR template, with the same things @jnurthen, see above -- always funny when you have the same exact idea has someone in the past.

should we make them all the same though? for familiarity? or is that silly and unnecessary? and I should just close this PR?

scottaohara:
@spectranaut just pulled in your wording into the original template.  so we're good here.

PR #442, opened 2022-10-19
update s element mapping

### LINKED ISSUE #416: Consider mapping <s> element to role=deletion
follow on from https://github.com/w3c/html-aam/issues/373

Filing as needing to further discuss the idea of mapping the `<s>` element to the deletion role.  If this is done, then there would likely be some clarifications/editorial for the deletion role definition in the ARIA spec.

scottaohara:
closes #416 

revises the `<s>` element to map to ARIA's `deletion` role, rather than having no implicit ARIA mapping. Introducing this mapping would allow for browsers to expose this role, and thus AT to better communicate that the content is marked as no longer relevant.  Presently authors need to employ visually hidden content to programmatically convey what the element visually represents with its default UA CSS strike-through.

## Implementation requests:
   * WebKit: [ISSUE: 246745](https://bugs.webkit.org/show_bug.cgi?id=246745)
   * Gecko: [ISSUE: 1796178](https://bugzilla.mozilla.org/show_bug.cgi?id=1796178)
   * Blink: [ISSUE: 1376572](https://bugs.chromium.org/p/chromium/issues/detail?id=1376572)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/html-aam/pull/442.html" title="Last updated on Dec 12, 2022, 5:40 PM UTC (599960a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/html-aam/442/e3a102a...599960a.html" title="Last updated on Dec 12, 2022, 5:40 PM UTC (599960a)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2659,33 +2659,13 @@
             </tr>
             <tr tabindex="-1" id="el-s">
               <th><a data-cite="HTML">`s`</a></th>
-              <td class="aria">No corresponding role</td>
-              <td class="ia2">
-                <div class="general">No accessible object.</div>
-                <div class="properties">
-                  <span class="type">Text attributes:</span> `text-line-through-style:solid` on the text container
-                </div>
+              <td class="aria">
+                <a class="core-mapping" href="#role-map-deletion">`deletion`</a> role
               </td>
-              <td class="uia">
-                <div class="general">No accessible object. Styles used are exposed by UIA text attribute identifiers of the `TextRange` Control Pattern implemented on a parent accessible object.
-                </div>
-              </td>
-              <td class="atk">
-                <div class="general">No accessible object. Exposed as
-                  "text-line-through-style:solid" text attribute on the text container.
-                </div>
-              </td>
-              <td class="ax">
-                <div class="role">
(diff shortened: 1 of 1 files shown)


jcsteh:
> Presently authors need to employ visually hidden content to programmatically convey what the element visually represents with its default UA CSS strike-through.

Just for the sake of clarity, authors could also use `role="deletion"` rather than visually hidden content.

scottaohara:
Thank you for the clarifying comment @jcsteh. Yes, my comment was meant to allude to practices that have been promoted prior to that role’s wider implementation, and the fact that there is a “semantic” difference between the html del and s elements. I have a minor companion Pr to edit the aria spec so that the deletion role definition covers both use cases for the different html elements.

spectranaut (approved):
This looks good to me, btw!

scottaohara:
I'm going to merge this since it's been picked up by @spectranaut for implementation in chromium, and Firefox has marked the bug as an enhancement.

cookiecrook (commented):
LGTM. Better late than never.

PR #448, opened 2022-12-08
Clarification: label element and for attribute

scottaohara:
This PR clarifies some of the wording in the `label` element mappings where, for instance, UIA was written in a way that was specifically calling out labelling the `input` element - but there are a number of labelable elements that are not `input`.  Rather than say all of them or even say 'form controls' which is also not necessarily accurate, instead mention and link to the HTML spec's definition and listing of labelable elements.

The `label[for]` attribute was updated to consistently indicate that it's used for accname by all the platforms.  UIA's 'not mapped' text was replaced with a version of the text that existed in the `label` element's UIA mapping.  MSAA/IA2 mappings were updated to reference the labelable element section from HTML.

note: this work was done as I noticed these things needed clarifying while making other additions to #402 today.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/html-aam/pull/448.html" title="Last updated on Mar 22, 2023, 12:36 PM UTC (91a8276)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/html-aam/448/472383b...91a8276.html" title="Last updated on Mar 22, 2023, 12:36 PM UTC (91a8276)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2101,7 +2101,7 @@
                 </div>
                 <div class="relations">
                   <span class="type">Relations:</span> 
-                  `IA2_RELATION_LABEL_FOR` with a <a href="#el-label">`label`</a> and referred <a data-cite="html/forms.html#category-label">labelable element</a>
+                  `IA2_RELATION_LABEL_FOR` with a <a data-cite="html/forms.html#category-label">labelable element</a>
                   that is child to the `label` or referred to by the `label` element's <a href="#att-for-label">`for`</a> attribute. 
                   The associated labelable element has `IA2_RELATION_LABELLED_BY` pointing to the `label`.
                 </div>
@@ -2112,13 +2112,15 @@
                 </div>
                 <div class="properties">
                   <span class="type">Other properties:</span>
-                  When the `label` element contains a <a data-cite="html/forms.html#category-label">labelable element</a>, the `LabeledBy` property for 
-                  the element points to the UIA element for the `label` element.
-                  <p>
+                  <div>
+                    When the `label` element contains a <a data-cite="html/forms.html#category-label">labelable element</a>, the `LabeledBy` property for 
+                    the element points to the UIA element for the `label` element.
+                  </div>
+                  <div>
                     When the `label` element has a <a href="#att-for-label">`for`</a> attribute referencing a  
                     <a data-cite="html/forms.html#category-label">labelable element</a>, the `LabeledBy` property for the referenced element points to 
                     the UIA element for the `label` element.
-                  </p>
+                  </div>
                 </div>
               </td>
               <td class="atk">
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -2100,7 +2100,10 @@ <h3>HTML Element Role Mappings</h3>
                   <span class="type">Roles:</span> `ROLE_SYSTEM_STATICTEXT`; `IA2_ROLE_LABEL`
                 </div>
                 <div class="relations">
-                  <span class="type">Relations:</span> `IA2_RELATION_LABEL_FOR` with a form control that is child to the `label` or referred to by the `label` element's <a href="#att-for-label">`for`</a> attribute. The associated form element has `IA2_RELATION_LABELLED_BY` pointing to the `label`.
+                  <span class="type">Relations:</span> 
+                  `IA2_RELATION_LABEL_FOR` with a <a href="#el-label">`label`</a> and referred <a data-cite="html/forms.html#category-label">labelable element</a>
```
spectranaut: I think I like the originally wording a little better!

you wrote:
* `IA2_RELATION_LABEL_FOR` with a `label` and referred `labelable element` that is a child to the `label` or referred to by the `label` elements `for` attribute

I think maybe you just want:
* `IA2_RELATION_LABEL_FOR` with a `labelable element` that is a child to the `label` or referred to by the `label` elements `for` attribute

index.html
```diff
@@ -2119,7 +2127,8 @@ <h3>HTML Element Role Mappings</h3>
                 </div>
                 <div class="relations">
                   <span class="type">Relations: </span>
-                  `ATK_RELATION_LABEL_FOR` for a child form element or form element referred by <a href="#att-for-label">`for`</a> attribute. Note, related form element provides `ATK_RELATION_LABELLED_BY` pointing to the label.
+                  `ATK_RELATION_LABEL_FOR` for a child form element or form element referred by <a href="#att-for-label">`for`</a> attribute.
```
spectranaut: did you miss this one -- should this also be `labelable elements` not form element? And shouldn't we care when there is a `labelable element` contains in the `label` element here, as well?

index.html
```diff
@@ -2109,8 +2112,13 @@ <h3>HTML Element Role Mappings</h3>
                 </div>
                 <div class="properties">
                   <span class="type">Other properties:</span>
-                  When the `label` element contains an `input` element, the `LabeledBy` property for the `input` element points to the UIA element for the `label` element.
-                  <p>When the `label` element has a <a href="#att-for-label">`for`</a> attribute referencing another element, the `LabeledBy` property for the referenced element points to the UIA element for the `label` element.</p>
+                  When the `label` element contains a <a data-cite="html/forms.html#category-label">labelable element</a>, the `LabeledBy` property for 
+                  the element points to the UIA element for the `label` element.
```
spectranaut: nit: it's weird looking that this is also not in a `<p>`, maybe it should be?

spectranaut (changes requested):
You know I think its a bit weird that for the label element description,  UIA + IA2 have the same "logic" but its worded differently, if that makes sense. And if my comment is correct on ATK, include that as well.

Like you could just choose the IA2 or UIA description and swap out "property" or "relation" with "relation" or "property" and the name of the relation or property, etc. Otherwise it's kind of weird because if you are referring to this spec you might thing the differences are more complicated when they really are the same concepts!

For example, UIA could be:
* `LabeledBy` attribute points to a `labelable element` that is a child to the `label` or referred to by the `label` elements `for` attribute

index.html
scottaohara: yes i think i messed this one up.  thank you!

index.html
scottaohara: there are a bunch of markup bits that need to be fixed in this spec.  The use of `<p>` here would introduce some extra spacing that other cells don't have.  Rather, I'll wrap these each in divs so they're consistent and not introduce variants in the styles.

scottaohara:
re-requesting review @spectranaut @jnurthen @adampage

adampage (approved):
This is a helpful clarification. 🚀

index.html
```diff
@@ -2111,17 +2114,26 @@ <h3>HTML Element Role Mappings</h3>
                 </div>
                 <div class="properties">
                   <span class="type">Other properties:</span>
```
```suggestion
                  <span class="type">Relations:</span> 
```

index.html
```diff
@@ -2111,17 +2114,26 @@ <h3>HTML Element Role Mappings</h3>
                 </div>
                 <div class="properties">
```
```suggestion
                <div>
```

PR #451, opened 2022-12-13
clarify img mappings

### LINKED ISSUE #439: Should an img with no src be mapped to role=img?
Should the following  (non-conforming) `img` element without `src` and without `alt` map to `role=img`

```
<img/>
```

Chrome: nothing rendered on screen, but exposed as role=img
Firefox: nothing rendered on screen, but exposed as role=img
Safari: nothing rendered on screen, not in accessibility tree

The HTML standard and html-aam seem to be at odds here::

- html-aam maps all img elements without alt to role=img (even if they have no src)
- the HTML spec says an img with no alt and no src represents nothing (i.e. has no semantics)

The HTML spec says:

>If the [src](https://html.spec.whatwg.org/multipage/embedded-content.html#attr-img-src) attribute is not set and either the [alt](https://html.spec.whatwg.org/multipage/embedded-content.html#attr-img-alt) attribute is set to the empty string or the [alt](https://html.spec.whatwg.org/multipage/embedded-content.html#attr-img-alt) attribute is not set at all
The element [represents](https://html.spec.whatwg.org/multipage/dom.html#represents) nothing.

https://html.spec.whatwg.org/multipage/embedded-content.html#the-img-element

The HTML spec defines "represents nothing" as meaning the element has no semantics.

This came up in the ACT Rules face-to-face meeting. Usually an `img` with no accessible name would be a SC 1.1.1 failure, but this looks hard to justify for `` since there's no non-text content because there's no src.

Does it make sense to add an element without content, and that's never rendered, to the accessibility tree?

scottaohara:
closes #439

This PR provides more detailed conditions for when an `img` element is mapped to the `img` role. 

This was originally brought up to address what should happen when an `img` lacks or has been given an `src` attribute set to the empty string, as HTML defines such instances as "representing nothing".  However, in trying to just add in that condition, it made me realize the mappings were lacking additional nuance to how the element is exposed re: the existence of alternative naming mechanisms besides just `alt`.

I have the following examples to fit these conditions:
```html
Presentation role:

<img src> 
<img srcset> 
<img src srcset> 
<img alt>
<img src alt>
<img srcset alt>
<img src srcset alt>
<img src=foo.jpg alt>
<img src=foo.jpg srcset="..." alt>
<img src=foo.jpg alt title=foo>
<img srcset=foo.jpg alt title=foo>
<img src=foo.jpg srcset=... alt title=foo>

Image role:

<img alt=foo>
<img src=foo.jpg>
<img title=foo>
<img src alt=foo>
<img src title=foo>
<img aria-label=foo>
<img aria-labelledby=f><span id=f>foo</span>
<img src aria-label=foo>
<img src aria-labelledby=f><span id=f>foo</span>
<img src alt aria-label=foo>
<img src alt aria-labelledby=f><span id=f>foo</span>
<img src=foo.jpg alt aria-label=foo>
<img src=foo.jpg alt aria-labelledby=f><span id=f>foo</span>
```
Question on what to do with `src=" "`.  I'd assume that too should be treated like `src` and `src=""`.  So maybe this PR also needs to call out src with a whitespace value?

## Implementation

* WPT tests: [LINK]()
   * WebKit: [ISSUE]()
   * Gecko: [ISSUE]()
   * Blink: [ISSUE]()

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/html-aam/pull/451.html" title="Last updated on Oct 3, 2023, 3:56 PM UTC (27bb2f5)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/html-aam/451/5856839...27bb2f5.html" title="Last updated on Oct 3, 2023, 3:56 PM UTC (27bb2f5)">Diff</a>


index.html
```diff
@@ -1322,31 +1322,47 @@ <h3>HTML Element Role Mappings</h3>
               <th>
                 <a data-cite="HTML">`img`</a>
               </th>
-              <td class="aria"><a class="core-mapping" href="#role-map-img">`img`</a> role</td>
+              <td class="aria">
+                <div class="role">
+                  <div>If element has a non-empty `src` attribute, an <a data-cite="accname-1.2/#dfn-accessible-name">accessible name</a>, or both, 
+                    then map to the <a class="core-mapping" href="#role-map-img">`img`</a> role.</div>
+                  <div>Otherwise, if the element has no `src` attribute, or it is set to the empty string, and has no `alt` attribute or
```
```suggestion
                  <div>Otherwise, if the element has no `src` or `srcset` attribute, or the attribute or attributes are set to the empty string, and has no `alt` attribute or 
```

index.html
```diff
@@ -1322,31 +1322,47 @@ <h3>HTML Element Role Mappings</h3>
...
+                </div>
+              </td>
               <td class="ia2"><div class="general">Use WAI-ARIA mapping</div></td>
               <td class="uia"><div class="general">Use WAI-ARIA mapping</div></td>
               <td class="atk"><div class="general">Use WAI-ARIA mapping</div></td>
               <td class="ax"><div class="general">Use WAI-ARIA mapping</div></td>
               <!-- <td class="naming"></td> -->
-              <td class="comments"></td>
+              <td class="comments">
+                <p>Authors can specify the `src` attribute to be the empty string by either marking it up as `src=""` or `src` with no value.</p>
```
```suggestion
```

index.html
```diff
@@ -1322,31 +1322,47 @@ <h3>HTML Element Role Mappings</h3>
...
+                <p>An `img` with a valid `src` and no `alt` attribute or <a data-cite="accname-1.2/#dfn-accessible-name">accessible name</a> will still map to
```
```suggestion
                <p>An `img` with a valid `src`, `srcset`, or both and no `alt` attribute or <a data-cite="accname-1.2/#dfn-accessible-name">accessible name</a> will still map to 
```

scottaohara:
Added updates to this PR to indicate that `srcset` should also be used in determining which role the `img` element maps to - particularly due to https://github.com/whatwg/html/issues/9181 where `src` is being proposed to be optional if an `srcset` attribute is present on the `img`.

scottaohara:
@jnurthen @smhigley ping for review on this so i can make bugs and get this finally merged/resolved

scottaohara:
@cookiecrook would apprecaite your thoughts on this clarification / if I should just create a wpt test for this to help move it along?

index.html
```diff
@@ -2964,12 +2964,23 @@ <h4 id=el-img>`img`</h4>
     <tr>
       <th>[[wai-aria-1.2]]</th>
       <td>
-        <a class="core-mapping" href="#role-map-image">`image`</a>
-        or  <a class="core-mapping" href="#role-map-img">`img`</a>role
+        <div>
+          If element has a non-empty `src` or `srcset` attribute, an <a data-cite="accname-1.2/#dfn-accessible-name">accessible name</a>, or both,
```
cookiecrook, line 3219: here you say "non-empty" attrs. below you say "valid" attrs. Is there a difference? Shouldn't a broken image still be conveyed as an image? Does an image referencing an SVG `src` containing plain text count still count as an image?

Perhaps more importantly, is there a problem with the way images are exposed now in the edge cases you listed like `?

cookiecrook:
> @cookiecrook would appreciate your thoughts on this clarification / if I should just create a wpt test for this to help move it along?

It couldn't hurt. Well I guess it could hurt, if your expectations are wrong, and the resolving engineer doesn't question it.

I would start with a new test file of subtests that include all these variants. If the engines agree on the output, or you are certain it's right check them in. If you aren't sure in the case of one or more subtests, leave it commented or otherwise not running, and file an issue, referencing the issue in the comment.

I did something similar recently with:
- https://github.com/w3c/accname/issues/209
- https://github.com/web-platform-tests/wpt/blob/master/accname/name/comp_labelledby.html#L41

index.html
```diff
@@ -2996,6 +3007,8 @@ <h4 id=el-img>`img`</h4>
     <tr>
       <th>Comments</th>
       <td>
+        <p>An `img` with a valid `src`, `srcset`, or both and no `alt` attribute or <a data-cite="accname-1.2/#dfn-accessible-name">accessible name</a> will still map to
```
smhigley: +1 to @cookiecrook's comment about valid vs. empty not being the same. My personal leaning would be to drop "valid" and use "non-empty" here.

index.html
```diff
@@ -3051,7 +3066,10 @@ <h4 id=el-img-empty-alt>`img` <span class="el-context">(`alt`attribute value is
     <!-- <th><a href="#accessible-name-and-description-computation">Naming Algorithm</a></th> -->
     <tr>
       <th>Comments</th>
-      <td>
+      <td class="comments">
+        <p>Authors can specify the `alt` attribute to be the empty string by either marking it up as `alt=""` or `alt` with no value.</p>
+        <p>If an `img` has been provided an `alt=""`, but also has an `aria-label` or `aria-labelledby` attribute specified which returns an
```
smhigley, line 3331: Not entirely related to this PR, but I wonder if this comment should be in the first row rather than in a note

smhigley (approved):
LGTM, pending that one empty vs. valid comment from James :)

index.html
```suggestion
        <p>An `img` with a non-empty `src`, `srcset`, or both and no `alt` attribute or <a data-cite="accname-1.2/#dfn-accessible-name">accessible name</a> will still map to 
```

index.html
scottaohara: hmm. i see what you're saying...but then i'd wonder why not just merge the two image tables together. i dunno, i don't want to address this in this PR, but i do want to think on it a bit more.  i'll file a different issue for this.

edit: issue filed so resolving this question for now.

index.html
scottaohara: updated.  thanks!

rahimabdi:
- https://github.com/w3c/aria/pull/2221

rahimabdi:
- https://github.com/w3c/aria/pull/2221

PR #454, opened 2023-01-10
Define concept of minimum role

scottaohara:
Worked on this with @aleventhal as this concept is necessary for some current HTML features (as identified in the attribute mapping table in this PR) as well as future proposed features, such as `popover`.

## Implementation

* WPT tests: 
   * WebKit:
   * Gecko: 
   * Blink: https://bugs.chromium.org/p/chromium/issues/detail?id=1474047
* Tests
   * https://github.com/web-platform-tests/wpt/pull/44934

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/html-aam/pull/454.html" title="Last updated on Jun 17, 2024, 10:49 PM UTC (fb66dfd)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/html-aam/454/e644368...fb66dfd.html" title="Last updated on Jun 17, 2024, 10:49 PM UTC (fb66dfd)">Diff</a>


index.html
```diff
@@ -215,32 +215,65 @@ <h3>Exposing HTML Features That Do Not Directly Map to Accessibility APIs</h3>
       </p>
     </section>
     <section>
-      <h3>HTML Element Role Mappings</h3>
+      <h3 id="minimum-role">Exposing HTML Features That Require a Minimum Role</h3>
+      <p>
+        A <dfn>minimum role</dfn> is the equivalent WAI-ARIA role an element will map to when an element has no more specific implicit role. This can help ensure that users of assistive technologies get the best possible experience for commonly-used and valid HTML markup where otherwise a role would not be exposed.
+      </p>
+      <p>
+        A minimum role is provided under the following conditions:
```
aleventhal: We should probably say, "when -all- of the following conditions are true:"

index.html
```diff
@@ -215,32 +215,65 @@ <h3>Exposing HTML Features That Do Not Directly Map to Accessibility APIs</h3>
...
+      </p>
       <ul>
-        <li>HTML elements with implicit WAI-ARIA role semantics MUST be mapped to platform <a class="termref">accessibility APIs</a> according to the identified WAI-ARIA role mapping as defined in the [[core-aam-1.2]] specification.</li>
```
aleventhal: Looks like this got moved, but not changed?

index.html
```diff
@@ -3678,7 +3713,11 @@ <h3>HTML Attribute State and Property Mappings</h3>
                 <div class="general">Use WAI-ARIA mapping</div>
```
aleventhal: Unrelated, but looks like there is a capitalization error for AXTextArea

index.html
```diff
@@ -3678,7 +3713,11 @@ <h3>HTML Attribute State and Property Mappings</h3>
...
               </td>
               <td class="comments">
-                If the element has the `contenteditable` attribute  and `aria-readonly="true"`, User Agents MUST expose only the `contenteditable` state.
+                <p>If the element has the `contenteditable` attribute  and `aria-readonly="true"`, User Agents MUST expose only the `contenteditable` state.</p>
+                <p>
+                  If the attribute exposes or inherits the `true` state, it provides a <a href="termref">minimum role</a> of <a class="core-mapping" href="#role-map-group">`group`</a>.
+                </p>
+                <!-- note: when/if contenteditable=plain-text becomes a standard, then this would have a minimum mapping of textbox -->
```
aleventhal: Or it is called plaintext-only?

scottaohara: yes it is.  will correct to plaintext-only in the comment

index.html
scottaohara: will do

index.html
scottaohara: yeh, the inclusion of the above section seems to have caused all the following lines to be marked as removed/edited.

index.html
scottaohara: i might as well fix it.

index.html
```diff
@@ -215,32 +215,65 @@ <h3>Exposing HTML Features That Do Not Directly Map to Accessibility APIs</h3>
...
+        <li>"Not mapped" means the element is not exposed via an <a class="termref">accessibility API</a>. Commonly, an element is not mapped because it's not displayed as part of the user interface. However, authors can force some of these elements to be rendered. For instance, by overriding user agent styles to render elements that would have been otherwise set to `display: none`. In these cases, the user agent SHOULD map such elements to the role of <a class="core-mapping" href="#role-map-generic">`generic`</a>, unless other HTML features have been specified which would require a more specific <a>minimum role</a> to be exposed.</li>
+        <li>Where an element is indicated as having <q>No corresponding (WAI-ARIA) role</q>, or is mapped to the <a class="core-mapping" href="#role-map-generic">`generic`</a> role, user agents MUST NOT expose the <a class="core-mapping" href="#ariaRoleDescription">`aria-roledescription`</a> property value in the <a class="termref">accessibility tree</a> unless the element has an explicit, conforming `role` attribute value which [[WAI-ARIA-1.2]] does not prohibit the use of `aria-roledescription` be used with.</li>
```
adampage:
```suggestion
        <li>Where an element is indicated as having <q>No corresponding (WAI-ARIA) role</q>, or is mapped to the <a class="core-mapping" href="#role-map-generic">`generic`</a> role, user agents MUST NOT expose the <a class="core-mapping" href="#ariaRoleDescription">`aria-roledescription`</a> property value in the <a class="termref">accessibility tree</a> unless the element has an explicit, conforming `role` attribute value for which [[WAI-ARIA-1.2]] does not prohibit the use of `aria-roledescription`.</li>
```

I think this tweak preserves the meaning and is more quickly understood?

index.html
```diff
@@ -3674,11 +3709,15 @@ <h3>HTML Attribute State and Property Mappings</h3>
               </td>
               <td class="ax">
                 <span class="type">Role:</span>
-                <a href="#el-textarea">AXtextArea</a>
+                <a href="#el-textarea">AXTextArea</a>
                 <div class="general">Use WAI-ARIA mapping</div>
               </td>
               <td class="comments">
-                If the element has the `contenteditable` attribute  and `aria-readonly="true"`, User Agents MUST expose only the `contenteditable` state.
+                <p>If the element has the `contenteditable` attribute  and `aria-readonly="true"`, User Agents MUST expose only the `contenteditable` state.</p>
```
adampage:
```suggestion
                <p>If the element has the `contenteditable` attribute and `aria-readonly="true"`, User Agents MUST expose only the `contenteditable` state.</p>
```

I apologize for this objectively useless suggestion.

adampage (approved):
I have no feedback of substance 😅, just a couple proofing suggestions.

index.html
```diff
@@ -215,32 +215,65 @@ <h3>Exposing HTML Features That Do Not Directly Map to Accessibility APIs</h3>
       </p>
     </section>
     <section>
-      <h3>HTML Element Role Mappings</h3>
+      <h3 id="minimum-role">Exposing HTML Features That Require a Minimum Role</h3>
+      <p>
+        A <dfn>minimum role</dfn> is the equivalent WAI-ARIA role an element will map to when an element has no more specific implicit role. This can help ensure that users of assistive technologies get the best possible experience for commonly-used and valid HTML markup where otherwise a role would not be exposed.
```
spectranaut: The way the first is worded, it seems to me like there is a concept of a hierarchy of role specificity (more of less specific), when I think really there is just no explicit role/generic and everything else?

It's really a nit, you can take or leave, but I'd write:

```suggestion
        A <dfn>minimum role</dfn> is the equivalent WAI-ARIA role an element will map to when an element does not have a specified or implicit non-generic role. This can help ensure that users of assistive technologies get the best possible experience for commonly-used and valid HTML markup where otherwise a role would not be exposed.
```

index.html
```diff
@@ -215,32 +215,65 @@ <h3>Exposing HTML Features That Do Not Directly Map to Accessibility APIs</h3>
...
         <li>
-          <strong>AXAPI:</strong>
-          <ul>
-            <li>User agents should return a user-presentable, localized string value for the Mac Accessibility AXRoleDescription.</li>
-          </ul>
+          Some HTML elements expose implicit WAI-ARIA roles depending on whether they have been provided an <a data-cite="accname-1.2/#dfn-accessible-name">accessible name</a>. How an element participates in the computation of its own or another element's <a data-cite="accname-1.2/#dfn-accessible-name">accessible name</a> and/or <a data-cite="accname-1.2/#dfn-accessible-description">accessible description</a> is described in the <a href="#accessible-name-and-description-computation">Accessible Name and Description Computation</a> section of this document.
         </li>
+        <li>HTML elements with no mapping, or an implicit mapping of `generic`
```
spectranaut: This looks like a mistaken fragment!
```suggestion
```

index.html
```diff
@@ -215,32 +215,65 @@ <h3>Exposing HTML Features That Do Not Directly Map to Accessibility APIs</h3>
...
       </ul>
```
```suggestion
```

spectranaut (approved):
This is an approval caveat remove the sentence fragment, at least!

spectranaut:
Maybe it's time to open issues on the browsers with this change?

index.html
scottaohara: not sure what this was requesting.  the removal of the /ul, that doesn't seem right?  or deleting the line break that was in place after it (i did that in another commit), so i'm going to resolve this since there doesn't appear to be any broken markup when validating it

spectranaut: whoops I made a mistake.

index.html
scottaohara: i'm leaning to leave as is, unless you hae a suggestion on how to indicate the list of conditions that follow are essentially what is clarifying what a more specific role means.

additionally, added a commit to clarify that an explicit role must also not be none, presentation or generic.

index.html
```diff
@@ -215,32 +215,64 @@ <h3>Exposing HTML Features That Do Not Directly Map to Accessibility APIs</h3>
       </p>
     </section>
     <section>
-      <h3>HTML Element Role Mappings</h3>
+      <h3 id="minimum-role">Exposing HTML Features That Require a Minimum Role</h3>
+      <p>
+        A <dfn>minimum role</dfn> is the equivalent WAI-ARIA role an element will map to if the element has no more specific implicit role, e.g., a non-generic role. This can help ensure that users of assistive technologies get the best possible experience for commonly-used and valid HTML markup where otherwise a role would not be exposed.
```
```suggestion
        A <dfn>minimum role</dfn> is the equivalent WAI-ARIA role an element will map to if the element does not have a more specific implicit role, e.g., a non-generic role. This can help ensure that users of assistive technologies get the best possible experience for commonly-used and valid HTML markup where otherwise a role would not be exposed.
```

index.html
cookiecrook: Bah! Keep the unrelated changes out of the diff... We shouldn't screw up the blame traces, and we don't need merge conflicts, especially since we'll probably remove that entire line with #455

index.html
```diff
@@ -3674,11 +3708,15 @@ <h3>HTML Attribute State and Property Mappings</h3>
               </td>
               <td class="ax">
                 <span class="type">Role:</span>
-                <a href="#el-textarea">AXtextArea</a>
+                <a href="#el-textarea">AXTextArea</a>
```
cookiecrook: Keeping unrelated changes out of the diff... See next comment thread (applied to line 3678 for some reason) 

👀 Watching you two!
```suggestion
                <a href="#el-textarea">AXtextArea</a>
```

index.html
```diff
@@ -215,32 +215,64 @@ <h3>Exposing HTML Features That Do Not Directly Map to Accessibility APIs</h3>
...
+            </ul>
+          </li>
+          <li>
+            <strong>AXAPI:</strong>
+            <ul>
+              <li>User agents should return a user-presentable, localized string value for the Mac Accessibility AXRoleDescription.</li>
+            </ul>
+          </li>
+        </ul>
+      </section>
```
cookiecrook: Diff is difficult to read... Was there any change to "HTML Element Role Mappings" or is this just a move/reinsertion that GitHub isn't displaying correctly? I see it was removed above (rather, your new insertion was above this) and I will review that insertion separately.

cookiecrook (changes requested):
I see a few problems with this and suggest we schedule a deep dive. 

The first is that this doesn't seem to account for native-reserved roles, or roles that we can't yet or haven't yet mimicked with ARIA. For example, `<input type="date" autofocus>` or `<video tabindex="0">`… I think this change suggests UAs _**MUST**_ map both to `group`, even though it wouldn't be logical to do so. Even if ARIA eventually adds a role for native date pickers, `video` is likely to remain separate for a while, because there isn't a good method to support the video API that goes along with it.

The second issue is that ARIA's concept of group requires a much more explicit and purposeful author decision, IMHO, than the related concept in the AX API... As a reminder, `AXGroup` predates ARIA's `group` by almost a decade, and is used as a generic container. ARIA's `group`, as I understand it, represents some explicit logical grouping, so I am not certain it's a good change to force promotion of something like `<div contenteditable>` when the author may have added the relevant role and label to another nearby container. E.g.,

```html
<main aria-label="Edit post">
  <div contenteditable>
    Now I have an extraneous, unlabeled group inside my main, where I hadn't before.
  </div>
</main>
```

Update b/c I thought of another one. Does the sub-span in this one really need a minimum role?
```html
<main contenteditable="true">
    Obi-Wan never told you what happened to your father.
    He told me enough! He told me you killed him!
    <span contenteditable="false">No. I am your father.</span>
</div>
```

index.html
cookiecrook: Referenced "[next comment](https://github.com/w3c/html-aam/pull/454#discussion_r1091230177)" is hidden in an outdated thread, so repeating here. 

> We shouldn't screw up the blame traces, and we don't need merge conflicts, especially since we may eventually remove that line with #455

scottaohara:
@cookiecrook thank you for the review.

per your two comments:
>...that this doesn't seem to account for native-reserved roles...

it was definitely the intent to account for that, so happy to incorporate changes to make sure that's included / I'll have a think on other wording to make sure that's clear.

re:
>so I am not certain it's a good change to force promotion of something like `<div contenteditable>` when the author may have added the relevant role and label to another nearby container.

that's an interesting idea.  Though it makes me think back to past projects I worked on - one being a website builder where there were blocks of editable areas within predefined templates.  So a main could have 2 or 3 contenteditable divs within it.  Something to think about...

cookiecrook:
> per your two comments:

FYI Our comments crossed when I added a third to the list: `contenteditable="false"` or `contenteditable="plaintext-only"` override cascades on sub-sections may not need explicit roles... example above.

scottaohara:
@cookiecrook we had talked about `false` not triggering a need for a minimum role.  In the comments for contenteditable it indicates the minimum role only changes if the value is true.  I have a comment in the source code to mention what happens with `plaintext-only` when that is added to the standard.

index.html
```diff
@@ -215,32 +215,64 @@ <h3>Exposing HTML Features That Do Not Directly Map to Accessibility APIs</h3>
...
-        <li>Where an element is indicated as having &quot;No corresponding (WAI-ARIA) role&quot;, or is mapped to the <a class="core-mapping" href="#role-map-generic">`generic`</a> role, user agents MUST NOT expose the <a class="core-mapping" href="#ariaRoleDescription">`aria-roledescription`</a> property value in the <a class="termref">accessibility tree</a> unless the element has an explicit, conforming `role` attribute value which [[WAI-ARIA-1.2]] does not prohibit the use of `aria-roledescription`.</li>
         <li>
-          <strong>IAccessible2:</strong>
-          <ul>
-            <li>All elements with accessible objects should implement the IAccessible, IAccessible2 and IAccessible2_2 interfaces.</li>
-          </ul>
+          the element either has no implicit mapping or has an implicit `generic` role mapping,
```
```suggestion
          the element either has no implicit mapping or has an implicit `generic` role mapping, 
```

for this line, I'm looking for suggestions on how to call out that platform-specific roles would also remain.  e.g., a video has its own platform role, but not a specific ARIA role.  @cookiecrook ideas?

spectranaut:
We had a deep dive this morning, it seems the main conclusions were:
- to remove the minimum role of group from contenteditable, as there is already a history of browsers/ATs attemping to make contenteditable work and we can't be sure if making those elements "group" would make things better or worse.
- there was agreement that the minimum role of contenteditable=plaintextonly could be "textbox"
- there was quite a bit of discussion about whether or not "minimum role" was the right language (instead of "implicit role for this attribute", opponents say the "implicit role" has a specific meaning related to the tagname so sound cause confusion).  @mcking65, who thought it was unclear, committed to reviewing this PR in order to understand.
- It was decided not to use the language "fallback" because that implies that authors shouldn't rely on the minimum role, when in fact, the aim is to allow them to rely on the minimum role. If adding an attribute makes the semantic meaning of  the element clear, then you shouldn't need to also add a role. Therefore validators will have to be updated to not report errors when these attributes are used on elements with no implicit or explicit roles.

index.html
cookiecrook: From the discussion today, maybe make this a role selection algorithm, such as:

1. explicit role value e.g. role="switch checkbox" (note: in earlier ARIA discussions, checkbox is the "fallback role" if switch is not implemented in the browser, though I don't recall if we codified that term anywhere)
2. Implicit role (e.g. from HTML-AAM) or native implicit implementation role (like `<video>`)
3. If current "role" from above steps is none or generic, use "minimum implicit role" if defined in AAM.
4. no role? ignored.

cookiecrook:
> It was decided not to use the language "fallback" 

Also that term had been used to describe the fallback roles such as `role="switch checkbox"` where `checkbox` was the ARIA 1.0 fallback role if the engine did not yet support the ARIA 1.1 `switch` role.

scottaohara:
@cookiecrook, @aleventhal, @spectranaut 
I've taken a stab at both revising the wording of the current condition steps AND have tried making the steps into a role algorithm.  I'm open to any feedback to go with or improve either direction

 <p>
        Use the following steps to determine if an element needs to expose a minimum role:
      </p>
      <ol>
        <li>The element has valid specified WAI-ARIA role,</li>
        <li>The element has an implicit WAI-ARIA role, or native platform mapping role,</li>
        <li>If the above checks return a computed role of `generic` or `none` and the element has been provided an attribute which requires a minimum role mapping for the element, then provide the <a>minimum role</a> as defined by the specified attribute.</li>
        <li>If the element has multiple attributes specified which require a minimum role be exposed, prioritize exposing a non-`group` role, if one is defined.</li>
      </ol>
      <hr>
      <p>
        A minimum role is provided when all of the following conditions are true:
      </p>
      <ul>
        <li>
          the author has <strong>not</strong> specified a valid explicit WAI-ARIA role to the element, 
          <strong>or</strong> the specified role is either `none`, `presentation`, or `generic`,
        </li>
        <li>
          the element either has no implicit WAI-ARIA role or platform role mappings, or has an implicit `generic` role mapping, 
        </li>
        <li>
          and the author has specified properties which require a minimum role mapping for the element.
        </li>
      </ul>
      <p>
        The properties which require a minimum role mapping are specific global HTML attributes, identified in the 
        <a href="#attribute-mapping-table">Mappings of HTML attributes table</a>.
      </p>
      <p>
        When these conditions are met, the browser MUST expose an object using the mappings defined in CORE-AAM for the specified minimum role.  If the element has multiple attributes specified which require a minimum role be exposed, prioritize exposing a non-`group` role, if one is defined.
      </p>

aleventhal:
I like the plain bulleted version better, it reads much easier.

cookiecrook:
Are there ever circumstances where two attributes might require conflicting minimum roles? Is there a prioritization of roles or attributes? 

Spitballing as aria-checked for checkbox and aria-level for heading. Regardless if that’s a good example, how would you determine which minimum role wins?

Can extension specs define these minimums too?

cookiecrook:
Editorial: ARIA editorial guidelines prefer the word “attributes” over “properties” unless there is a substantive need to differentiate between state and property in that specific context.

scottaohara:
i pulled in the edits for the bulleted list version into my latest update to the PR.  

swapped out references to properties for attributes, per @cookiecrook's feedback.

Re: the question about what to do if different `aria-*` attributes are used and how to determine a minimum role from that, or extension specs.  That goes beyond what we had planned for here, as this was limited to HTML features only.  I would imagine that if that were to be defined, it'd need to be defined in the ARIA / extension spec, as I don't think HTML AAM would be the place for that.  Also, it'd likely be in the handling author error section, and would likely need to account for not only the attributes used, but also account for the a11y-parent of the element in question.  e.g., `<div aria-checked=true>` is that a radio, checkbox, treeitem, menuitemcheckbox/radio?

re:
>Are there ever circumstances where two attributes might require conflicting minimum roles? Is there a prioritization of roles or attributes?

The example I was thinking of when i added the text alluding to this is that one could have something like `<div tabindex=0 contenteditable=plaintext-only>` where tabindex would provide a group role, but this contenteditable value we were considering as a role=textbox, since other rich editing is not allowed for this type of contenteditable.

cookiecrook:
@scottaohara wrote:
> The example I was thinking of when i added the text alluding to this is that one could have something like `<div tabindex=0 contenteditable=plaintext-only>` where tabindex would provide a group role, but this contenteditable value we were considering as a role=textbox, since other rich editing is not allowed for this type of contenteditable.

Doesn't that mean a cascade/hierarchy defining that `textbox` is greater than `group` would need to be added for the context of minimum role?

scottaohara:
@cookiecrook ayup.  While I realize it's not very descriptive at this point (as everything is either group or the one instance of textbox), i did add in the text:
>If the element has multiple attributes specified which require a minimum role be exposed, prioritize the non-`group` role.

If that's not suitable for now, we can write up a list overtly stating the priority would be first textbox, then group.  I've been trying to think up other potential roles that could slot in here if this were to get more complex, but i'm running a blank right now.

cookiecrook:
> prioritize the non-group role.

Maybe "the more specific role in the ARIA taxonomy"?

index.html
```diff
@@ -215,32 +215,68 @@ <h3>Exposing HTML Features That Do Not Directly Map to Accessibility APIs</h3>
...
+          and the author has specified attributes which require a minimum role mapping for the element.
         </li>
       </ul>
+      <p>
+        The <a href="#attribute-mapping-table">Mappings of HTML attributes table</a> identifies the specific global attributes which would require
+        an element map to a minimum role.
+      </p>
+      <p>
+        When these conditions are met, the browser MUST expose an object using the mappings defined in CORE-AAM for the specified minimum role.
+        If the element has multiple attributes specified which require a minimum role be exposed, prioritize the non-`group` role.
```
```suggestion
        If the element has multiple attributes specified which require a minimum role be exposed, prioritize the more specific role in the ARIA taxonomy.
```

scottaohara:
@cookiecrook thanks for the suggestion. adjusted to include that.

cookiecrook (approved):

aleventhal (approved):

aleventhal:
We have this ready to land, but I think we should double check with screen reader vendors because it's a big change. It's probably better but I want to be safe. For example, `<div tabindex=0>` will now be a group, which could impact a lot of sites.

@jcsteh have you looked at it?

stevefaulkner:
> We have this ready to land, but I think we should double check with screen reader vendors because it's a big change. It's probably better but I want to be safe. For example, `<div tabindex=0>` will now be a group, which could impact a lot of sites.
> 
> @jcsteh have you looked at it?

given that `role=group` is generally ignored unless it has an accessible name this should not be a problem should it? Will ask Glen Gordon and Brett Lewis from Freedom Sci for input.

index.html
```diff
@@ -5643,7 +5687,9 @@ <h3>HTML Attribute State and Property Mappings</h3>
               <td class="uia"><div class="general">Use WAI-ARIA mapping</div></td>
               <td class="atk"><div class="general">Use WAI-ARIA mapping</div></td>
               <td class="ax"><div class="general">Use WAI-ARIA mapping</div></td>
-              <td class="comments"></td>
+              <td class="comments">
+                Provides a <a href="termref">minimum role</a> of <a class="core-mapping" href="#role-map-group">`group`</a>.
```
aleventhal: Hmm, I think we should only do this for tabindex=0. If we do it for tabindex=-1, it might create a lot of noise on web pages where tabindex=-1 is put on random elements for hard-to-explain reasons.

We ran into this when we used to repair missing names on `<div tabindex=-1>` by computing the name from contents. It caused too much noise.

index.html
```diff
@@ -5643,7 +5687,9 @@ <h3>HTML Attribute State and Property Mappings</h3>
               <td class="uia"><div class="general">Use WAI-ARIA mapping</div></td>
               <td class="atk"><div class="general">Use WAI-ARIA mapping</div></td>
               <td class="ax"><div class="general">Use WAI-ARIA mapping</div></td>
-              <td class="comments"></td>
+              <td class="comments">
```
aleventhal: Now that Hunseop Jeong has implemented this in https://chromium-review.googlesource.com/c/chromium/src/+/4794310 I have questions, because I realize we used to repair nameless occurances of `<div tabindex=0>` by computing name from contents. We will no longer be doing this, which could break things like:

```
<p>I don't know what I'm doing but <div tabindex="0" onclick="clicked()">Click me</div></p>
```

Should we care about that?

BrettLewisVispero: Firefox doesn't seem to add name from content to <div tabindex="0"> which matches the ARIA spec regarding generic role. I think having the behavior match for Chrome and Firefox will be a positive. 
For JAWS, this change will cause JAWS to announce "group" on focus, but I thihnk that is fine. In addition, if chrome no longer adds name from content to generic elements, the cases where focus moves inside these generic elements with names will be less chatty.

aleventhal:
Ok, what about in the virtual buffer, e.g. with weird <div tabindex=-1>
cases?
Will those be problematic?

index.html
scottaohara: there are kinda two different cases that should be considered here.

a generic element with a tabindex so that it's focusable or in the tab order.  those fit under what brett mentioned.

then there are the generic elements with a tabindex which also have a click/some-kind-of event attached.  and those are more likely to be things where maybe the name should be calculated to make up for the likely faux buttons being made (not saying they 'should' be buttons.  we don't know...)

aleventhal: I suppose even without any changes to this PR, the fake `<div tabindex=0 onclick>` button will be accessible because the user will be able to read the contents in the virtual buffer. It might not act as nice as a real button, but that's arguably good.

So my next question is whether <div tabindex=-1> should get the minimum group role, or whether we should just leave those alone.

scottaohara: i'd still say group.  any more than that would make a lot of assumptions that could make the content more difficult to read / could still be wrong in terms of the role.  e.g., maybe  it's supposed to be a link.  maybe it's not really supposed to be either... so stay with group to keep a consistent baseline

jcsteh:
I somehow missed this altogether, so no, i haven't looked at it. I'll endeavour to do so soonish.

jcsteh:
I'm trying to grasp exactly what this gives us that we didn't have before:

1. I thought the spec currently required that focusable elements were exposed in the tree, but it seems it doesn't. So, I guess this makes it clear that `<div tabindex="0">` should be exposed, for example. Is that a primary intention here?
2. As far as I know, accesskey is only useful on something that is focusable. So the accesskey is probably covered by the tabindex case.
3. Is there an advantage of using "group" over "generic"? From what I've seen, group is not normally a role that is focusable. Group is no worse than generic, I guess, but generic is what some browsers already expose in cases like these, so I'm wondering why the change here rather than reflecting what already happens in some cases.
4. I do understand that this might be necessary for future proposed features, so if we're just setting up groundwork here, that's fine. I'm just trying to understand the impact on current features first.

scottaohara:
hi @jcsteh, thanks for giving this a review, and yes, your final point is definitely the crux of this proposal - create consistency in how these and other future / semi-present (`popover`) global attributes are handled, so that the different browsers and AT can coalesce around how generic elements, which should generally be ignored, can be exposed to users / handled by authors.

Some more comments related to your other questions:
1. Yes, this is meant to not only explicitly identify that focusable elements need to be exposed, but help to standardize how.  E.g., moving focus to a generic div when using firefox/chrome produce no announcement of any role with JAWS/NVDA, right now.  But presently VoiceOver and Narrator will announce this same element as a 'group' (putting aside whether because the API is already exposing this as a group, or this is just how they presently decided to announce a generic which has been marked as important in the a11y tree).

2. regarding `accesskey` (but also `draggable`), agreed that these attributes will commonly be used on focusable elements - but since they're global attributes, they don't _have to be_.  

3. The primary advantage again comes down to consistency / match what is already happening in some cases.  Allowing AT to expose a 'group' role arguably being more meaningful than exposing a 'generic' role, or having the AT or browser come up with their own unique role description for a generic.  Standardizing around 'group' could then allow consistency in how such elements were named, or not, or how if they had a `title` attribute, or other globally allowed `aria-*` attributes which are allowed on generics but often don't' get exposed to users in an overt way (if at all), they could instead be exposed in the manner in which they are with the group role.

jcsteh:
>     1. Yes, this is meant to not only explicitly identify that focusable elements need to be exposed, but help to standardize how.  E.g., moving focus to a generic div when using firefox/chrome produce no announcement of any role with JAWS/NVDA, right now.

From what I can see, focusing a div produces exactly the same result as focusing a group with NVDA, except that a div reports "section" and a group reports "grouping".

>     2. regarding `accesskey` (but also `draggable`), agreed that these attributes will commonly be used on focusable elements - but since they're global attributes, they don't _have to be_.

accesskey is interesting in that it is global, but as far as I can tell, it isn't useful if the element to which it is applied is not focusable. Firefox and Chrome both currently expose accesskey on a non-focusable element via accessibility APIs, but pressing the access key does nothing. This raises the question of whether we should be exposing this at all in this case, since it's misleading to the user. Making that decision is probably out of scope here, but it does impact whether we bother having a minimum role in this case. I can't see a benefit given the above.

>     3. The primary advantage again comes down to consistency / match what is already happening in some cases.  Allowing AT to expose a 'group' role arguably being more meaningful than exposing a 'generic' role, or having the AT or browser come up with their own unique role description for a generic.  Standardizing around 'group' could then allow consistency in how such elements were named, or not, or how if they had a `title` attribute, or other globally allowed `aria-*` attributes which are allowed on generics but often don't' get exposed to users in an overt way (if at all), they could instead be exposed in the manner in which they are with the group role.

The problem here is that this assumes either that group is entirely intuitive as a fallback or that it is already being used like this, and I'm not convinced either of these is true. As I noted, unless I'm missing something, group was never really intended to get focus itself, but rather, to be a container for other focusable elements. In that sense, group is no better than generic because in either case, it's not being used for its intended semantic purpose.

jcsteh:
To put this another, more practical way, if a screen reader user tabs to something and just hears "group", they're not going to know what it is. That's no better to the current situation where the user might tab to something and hear just "section", which is also meaningless. Arguably, it's worse because "group" has existing semantic meaning. At least with generic, screen readers could potentially say "unknown element" or something like that, which would signal to the user that the screen reader doesn't have sufficient information about what this is.

BrettLewisVispero:
One thing to remember, in Chrome currently, focusable generic elements get their name from content. Group would at least make it clearer that the name must be provided by author.
(of course generic element shouldn’t have names at all).

scottaohara:
had a call with aaron and james today
agreed that accesskey can be dropped from this
no disagreement about introducing the concept
still not 100% on tabindex, but not against it - it's just a change that may create other issues to review (re: naming divs with title which are now groups because of tabindex=0)
can this be merged with the other work i'm doing for contextual role (proably?)

i need to work though if this should be merged, or can be revised into contextual role.  i'll work on that soon

aleventhal:
@scottaohara what's the update on this? Which attributes should we start with?

index.html
```diff
@@ -184,13 +184,49 @@ <h3>Exposing HTML Features That Do Not Directly Map to Accessibility APIs</h3>
...
+        </li>
+        <li>
+          the element either has no implicit WAI-ARIA role or platform role mappings, or it has an implicit `generic` or `none` computed role; 
+        </li>
+        <li>
+          and the author has specified attributes which require a minimum role mapping for the element.
+        </li>
+      </ul>
+      <p>
+        The <a href="#attribute-mapping-table">Mappings of HTML attributes table</a> identifies the specific global attributes which would require
```
```suggestion
        The <a href=html-attribute-state-and-property-mappings">HTML Attribute State and Property Mappings</a> section identifies the specific global attributes which would require
```

index.html
```diff
@@ -184,13 +184,49 @@ <h3>Exposing HTML Features That Do Not Directly Map to Accessibility APIs</h3>
...
+        </li>
+        <li>
+          the element either has no implicit WAI-ARIA role or platform role mappings, or it has an implicit `generic` or `none` computed role; 
+        </li>
+        <li>
+          and the author has specified attributes which require a minimum role mapping for the element.
+        </li>
+      </ul>
+      <p>
+        The <a href=#html-attribute-state-and-property-mappings">HTML Attribute State and Property Mappings</a> section identifies the specific global attributes which would require
```
scottaohara: and missing quote...
```suggestion
        The <a href="#html-attribute-state-and-property-mappings">HTML Attribute State and Property Mappings</a> section identifies the specific global attributes which would require
```

scottaohara:
OK, finally got back to this and resolved all the merge conflicts.  The PR is much simpler to review now in the diff mode.

This now adds the minimum role concept, and to start we're indicating this will occur for the `draggable` and `autofocus` attributes.  

`tabindex` is presently not part of this PR (well, it's commented out for further investigation/discussion).

I plan on some editorial revising of this section with the contextual role PR (https://github.com/w3c/html-aam/pull/484) - but need this to merge first before I work on that anymore.

I think this finalizes the outstanding conversations for this PR (with any potential addition of tabindex being it's own topic for the future)

cc @aleventhal @cookiecrook @jcsteh @benbeaudry

aleventhal:
Still LGTM.

benbeaudry:
+1!

jcsteh (approved):

scottaohara:
merged via https://github.com/w3c/aria/pull/2220

PR #463, opened 2023-03-08
Clarify mappings for contenteditable

scottaohara:
Per discussion around the minimum role concept, and the WG deciding that contentedtiable should remain unmapped to any specific ARIA role when used on a generic.    per that decision #324 is closed, but the other clarifications that were to be a part of that PR are now in this one.

Additionally, there are other necessary updates needed for these mappings, as the previous ones did not get into the nuance of how to handle nested `contenteditable` attributes.

This PR does not include changes for `contenteditable=plaintext-only` (though there is some commented out text which _does_ mention this state).  Mappings for `plaintext-only` will be included in, or be a follow-on to #454.

`plaintext-only` landed in HTML - https://github.com/whatwg/html/pull/8275

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/html-aam/pull/463.html" title="Last updated on May 19, 2023, 12:56 PM UTC (f8bd96c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/html-aam/463/3c6988f...f8bd96c.html" title="Last updated on May 19, 2023, 12:56 PM UTC (f8bd96c)">Diff</a>


index.html
```diff
@@ -3658,39 +3658,56 @@ <h3>HTML Attribute State and Property Mappings</h3>
...
+                <!--<p>If in the `true`, `false` or `inherit` state: -->Not mapped<!--</p>
+                <p>If in the `plaintext-only` state: see comments</p>  -->
+              </td>
               <td class="ia2">
+                <p>If the element is in the editable state, the following mappings apply to the element and every nested accessible object with the 
+                exception of those which have been specified in the `false` state.</p>
                 <div class="states">
                   <span class="type">States:</span>
-                  `IA2_STATE_EDITABLE` on this and every nested text accessible object
+                  `IA2_STATE_EDITABLE` and `IA2_STATE_MULTI_LINE`
```
jcsteh, line 3658: Would this multi-line state be overridden by setting aria-multiline="false"? If so, do we need to state that explicitly or is that implied somewhere else?

jcsteh (commented):
Thanks. Looks good to me in general.

index.html
scottaohara: i was wondering if this might need a comment.  

multi-line functionality is not something that can easily be 'turned off' by a developer.  an element that has contenteditable will almost always allow for multiline editing, and even if someone were to use JS to disable character returns, or display: none breaks that were put in, a user could still likely paste their multi-line content into one of these almost single line contenteditables, and now there'd be a multiline contenteditable element again.

I think this is another scenario where the implicit semantics/functionality are strong enough that an ARIA attribute shouldn't be able to overwrite them.

happy for someone to inform me that I'm unaware of this one weird trick to for devs to achieve this, and thus allow the aria attribute to convey the editable as a single line. but i'd submit it'd likely be better to have this sort of feature implemented by HTML (allowing a contenteditable to a single line editable), and then an aria attribute wouldn't be needed.

jcsteh: [Today's episode of just how far web devs will go](https://stackoverflow.com/questions/6831482/contenteditable-single-line-input) :)

I guess that's fair, though you could equally argue that most things should be implemented in HTML and then ARIA wouldn't be needed. :) Still, this seems fine to me.

jcsteh (approved):

index.html
scottaohara: haha, yup. i've seen that exact stackoverflow, as well as others.  doesn't completely solve the issue, but yeh.

cookiecrook:
My review was requested in an ARIA meeting, but since the substantive changes are all related to Windows and Linux API, I'm removing myself as a reviewer.

PR #446, opened 2022-11-28
Adds proposed popover and related attributes

scottaohara:
See https://open-ui.org/components/popup.research.explainer and https://github.com/w3c/html-aam/wiki/HTML-Popover-Attribute-A11y-Proposal-(manual-and-auto)

Adds the following proposed attributes:
- `popover`
- `popoverhidetarget`
- `popovershowtarget`
- `popovertoggletarget`

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/html-aam/pull/446.html" title="Last updated on Feb 23, 2023, 6:35 PM UTC (c8b66f5)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/html-aam/446/0323bc1...c8b66f5.html" title="Last updated on Feb 23, 2023, 6:35 PM UTC (c8b66f5)">Diff</a>

PR #359, opened 2022-01-20
Change figure & figcaption accName computation

scottaohara:
resolves #325

This PR revises the mappings for `figure` and `figcaption` to remove the "labelled by" relations, changing them to "details for".
The accessible name computation for `figure` removes `figcaption`, a note is added to further reference this change.

- [x] need to update to UIA mapping (core aam issue: https://github.com/w3c/core-aam/issues/110)
- [x] do we want `figcaption` to map to `caption` (this is probably a 'yes', but want confirmation)? 
- [x] regardless of above, [`caption`](https://w3c.github.io/aria/#caption) will need to be updated in regards to its referencing `figure` and `figcaption` relationships (ARIA PR: https://github.com/w3c/aria/pull/1703)
- [x] [`aria-details`](https://w3c.github.io/aria/#aria-details) definition should be updated to reference this change (ARIA PR: https://github.com/w3c/aria/pull/1704)

Implementation commitment:

 * [ ] WebKit (https://bugs.webkit.org/show_bug.cgi?id=)
 * [x] Chromium 
 * [ ] Gecko (https://bugzilla.mozilla.org/show_bug.cgi?id=)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/html-aam/pull/359.html" title="Last updated on Dec 13, 2023, 1:35 PM UTC (2380d76)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/html-aam/359/ab38b66...2380d76.html" title="Last updated on Dec 13, 2023, 1:35 PM UTC (2380d76)">Diff</a>

PR #481, opened 2023-05-09
add/revise popover related attribute mappings

scottaohara:
this commit supersedes #446

TODO: finalize `popover` attribute mappings

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/html-aam/pull/481.html" title="Last updated on May 10, 2024, 5:23 PM UTC (fcdd9ac)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/html-aam/481/ff5bdf3...fcdd9ac.html" title="Last updated on May 10, 2024, 5:23 PM UTC (fcdd9ac)">Diff</a>


index.html
```diff
@@ -5009,6 +5009,63 @@ <h3>HTML Attribute State and Property Mappings</h3>
...
+              <td class="elements">
+              	<a data-cite="html/popover.html#attr-popovertargetaction">`button`</a>;
+                <a data-cite="html/popover.html#attr-popovertargetaction">`input type=button`</a>
+              </td>
+              <td class="aria"><div class="general">Not mapped</div></td>
+              <td class="ia2"><div class="general">Not mapped</div></td>
+              <td class="uia"><div class="general">Not mapped</div></td>
+              <td class="atk"><div class="general">Not mapped</div></td>
+              <td class="ax"><div class="general">Not mapped</div></td>
+              <td class="comments"></td>
```
scottaohara: question 1: should popovertargetaction=close be a signal to _not_ expose a details association with the button and the associated popover?  

question 2: if a button has a popovertargetaction=open or close, it will only perform this action and not serve as a toggle.  Should such buttons still be exposed as having an expand/collapse state ?  Or, should the control have a "disabled" state communicated when the associated popover is in the matchign state?  e.g., a button with popovertargetaction=close, where the popover is in the closed state, would be exposed as "collapsed" and "disabled" to indicate that this control cannot be used to open the popover.

index.html
```diff
@@ -5197,6 +5197,63 @@ <h3>HTML Attribute State and Property Mappings</h3>
...
+                </p>
+              </td>
+              <td class="ia2"><div class="general">Use WAI-ARIA mapping</div></td>
+              <td class="uia"><div class="general">Use WAI-ARIA mapping</div></td>
+              <td class="atk"><div class="general">Use WAI-ARIA mapping</div></td>
+              <td class="ax"><div class="general">Use WAI-ARIA mapping</div></td>
+              <td class="comments">
+                <p>
+                  If the associated element is not a direct sibling to the element with the `popovertarget` attribute in the DOM:
+                  create an <a class="core-mapping" href="#ariaDetalis">`aria-details`</a> relation with the associated element.
```
hidde:
```suggestion
                  create an <a class="core-mapping" href="#ariaDetails">`aria-details`</a> relation with the associated element.
```

(looks like this was a typo?)

index.html
scottaohara: resolving this and made a new issue.

index.html
scottaohara: ah thanks. fixed this in the larger conflict resolution i needed to do for the file

scottaohara:
made related follow up https://github.com/w3c/html-aam/issues/505

index.html
```diff
@@ -13593,6 +13593,193 @@ <h4 id=att-playsinline>`playsinline`</h4>
...
+            <a class="core-mapping" href="#ariaExpandedUndefined">`aria-expanded=undefined`</a>
+          </p>
+        </div>
+      </td>
+    </tr>
+    <tr>
+      <th><a href="https://msdn.microsoft.com/en-us/library/dd373608%28v=VS.85%29.aspx">MSAA</a> 
+        + <a href="http://accessibility.linuxfoundation.org/a11yspecs/ia2/docs/html/">IAccessible2</a></th>
+      <td>
+        ???
```
```suggestion
        <div class="general">Use WAI-ARIA mapping</div>
```

index.html
```diff
@@ -13594,6 +13594,207 @@ <h4 id=att-playsinline>`playsinline`</h4>
...
+      </td>
+    </tr>
+    <tr>
+      <th><a href="https://msdn.microsoft.com/en-us/library/dd373608%28v=VS.85%29.aspx">MSAA</a> + <a href="http://accessibility.linuxfoundation.org/a11yspecs/ia2/docs/html/">IAccessible2</a></th>
+      <td>
+        <span class="type">Relations:</span> `IA2_RELATION_DETAILS_FOR` points to invoking element
+        <div class="objattrs">
+          <span class="type">Object attributes:</span>
+          <ul>
+            <li>`ispopup: <value>` where `<value>` reflects the <a data-cite="html/popover.html#the-popover-attribute">popover</a> type</li>
```
```suggestion
            <li><code>ispopup: &lt;value></code> where `<value>` reflects the <a data-cite="html/popover.html#the-popover-attribute">popover</a> type</li>
```

index.html
```diff
@@ -13594,6 +13594,207 @@ <h4 id=att-playsinline>`playsinline`</h4>
...
+      </td>
+    </tr>
+    <tr>
+      <th><a href="https://msdn.microsoft.com/en-us/library/dd373608%28v=VS.85%29.aspx">MSAA</a> + <a href="http://accessibility.linuxfoundation.org/a11yspecs/ia2/docs/html/">IAccessible2</a></th>
+      <td>
+        <span class="type">Relations:</span> `IA2_RELATION_DETAILS_FOR` points to invoking element
+        <div class="objattrs">
+          <span class="type">Object attributes:</span>
+          <ul>
+            <li><code>ispopup: &lt;value></code> where `<value>` reflects the <a data-cite="html/popover.html#the-popover-attribute">popover</a> type</li>
```
```suggestion
            <li><code>ispopup: &lt;value></code> where <code>&lt;value></code> reflects the <a data-cite="html/popover.html#the-popover-attribute">popover</a> type</li>
```

index.html
```diff
@@ -13594,6 +13594,207 @@ <h4 id=att-playsinline>`playsinline`</h4>
...
+      </td>
+    </tr>
+    <tr>
+      <th><a href="https://gnome.pages.gitlab.gnome.org/atk/">ATK</a></th>
+      <td>
+        <span class="type">Relations:</span> `RELATION_DETAILS_FOR` points to invoking element
+        <div class="objattrs">
+          <span class="type">Object attributes:</span>
+          <ul>
+            <li>`ispopup: <value>` where `<value>` reflects the <a data-cite="html/popover.html#the-popover-attribute">popover</a> type</li>
```
```suggestion
            <li><code>ispopup: &lt;value></code> where <code>&lt;value></code> reflects the <a data-cite="html/popover.html#the-popover-attribute">popover</a> type</li>
```

index.html
```diff
@@ -13594,6 +13594,207 @@ <h4 id=att-playsinline>`playsinline`</h4>
...
+    </tr>
+    <tr>
+      <th><a href="https://msdn.microsoft.com/en-us/library/dd373608%28v=VS.85%29.aspx">MSAA</a> + <a href="http://accessibility.linuxfoundation.org/a11yspecs/ia2/docs/html/">IAccessible2</a></th>
+      <td>
+        <span class="type">Relations:</span> `IA2_RELATION_DETAILS_FOR` points to invoking element
+        <div class="objattrs">
+          <span class="type">Object attributes:</span>
+          <ul>
+            <li><code>ispopup: &lt;value></code> where <code>&lt;value></code> reflects the <a data-cite="html/popover.html#the-popover-attribute">popover</a> type</li>
+            <li>`details-roles:popover`</li>
```
jcsteh: If I understand correctly, details-roles should be exposed on the object with the details relation, not on the object with the detailsFor relation. This also seems to be how it is implemented in Chrome.

index.html
```diff
@@ -13594,6 +13594,207 @@ <h4 id=att-playsinline>`playsinline`</h4>
...
+    <tr>
+      <th>[[WAI-ARIA-1.2]]</th>
+      <td>
+        <div class="general">Not mapped</div>
+      </td>
+    </tr>
+    <tr>
+      <th><a href="https://msdn.microsoft.com/en-us/library/dd373608%28v=VS.85%29.aspx">MSAA</a> + <a href="http://accessibility.linuxfoundation.org/a11yspecs/ia2/docs/html/">IAccessible2</a></th>
+      <td>
+        <span class="type">Relations:</span> `IA2_RELATION_DETAILS_FOR` points to invoking element
```
jcsteh: It feels weird that we express this as a platform specific relation, but we express popovertarget's details relation as a cross-platform aria-details relation. Is this just because we have no way of expressing reverse relations for ARIA mappings and we want to be explicit here? I wonder whether it'd be better to link to a more general note about implicit reverse relations or something like that.

index.html
```diff
@@ -13594,6 +13594,207 @@ <h4 id=att-playsinline>`playsinline`</h4>
...
+      <td>
+        <div class="general">Use WAI-ARIA mapping</div>
+      </td>
+    </tr>
+    <tr>
+      <th>Comments</th>
+      <td>
+        <p>
+          User Agents MUST expose an <a class="core-mapping" href="#ariaDetails">`aria-details`</a> relation with the associated popover element
+          <strong>except</strong> under the following conditions:
```
jcsteh, line 14362: Is this "all of the following conditions" or "one of the following conditions"?

index.html
```diff
@@ -13594,6 +13594,207 @@ <h4 id=att-playsinline>`playsinline`</h4>
...
+        </p>
+        <ul>
+          <li>The element's `popovertargetaction` attribute value is <strong>not</strong> "`hide`"</li>
```
jcsteh: I don't understand this.

1. So we don't want to expose aria-details if the action is show? Is that because the target is hidden in that case, so we can't expose it? But what about toggle, which could be either show or hide depending on the state of the popover?
2. This doesn't seem to match Chrome's implementation, which exposes aria-details for `show` but not `hide` or `toggle`. I don't really understand why Chrome doesn't expose it for `toggle` when the target is shown, though.
3. Would it be better to express this in terms of whether the popover is showing, rather than the action? Or am I missing the intent here?

index.html
scottaohara, line 14362:
```suggestion
          <strong>except</strong> under either of the following conditions:
```

index.html
scottaohara: sorry for the confusion here.  this was worded differently before and the 'not' should have been removed in the previous update. 

i'm not sure why chrome isn't exposing the relation for toggle - it should per my understanding of how it was discussed this should work, unless the popover is a sibling to the invoking element.

```suggestion
          <li>The element's `popovertargetaction` attribute value is "`hide`"</li>
```

index.html
scottaohara: I'd be fine with either replacing the cross-platform aria-details relation with the specific platform mappings, or keeping the cross-platform relation and removing the platform specific and replacing with a note about the reverse relations.  
over in https://github.com/w3c/html-aam/pull/359 (which incase you missed, i pinged you for a review) we use the platform specific reverse relations for figure/figcaption.   So we can do that here too, or similarly change over in that PR.  

Whatever makes the most sense for those that need to read the spec is what I want to do, so just let me know which is preferred.

index.html
```diff
@@ -13594,6 +13594,208 @@ <h4 id=att-playsinline>`playsinline`</h4>
...
+    <tr>
+      <th>[[WAI-ARIA-1.2]]</th>
+      <td>
+        <div class="general">Not mapped</div>
+      </td>
+    </tr>
+    <tr>
+      <th><a href="https://msdn.microsoft.com/en-us/library/dd373608%28v=VS.85%29.aspx">MSAA</a> + <a href="http://accessibility.linuxfoundation.org/a11yspecs/ia2/docs/html/">IAccessible2</a></th>
+      <td>
+        <span class="type">Relations:</span> `IA2_RELATION_DETAILS_FOR` points to invoking element
```
jcsteh: I assume this shouldn't happen if we don't expose aria-details due to the invoking element being an immediate sibling or having popovertargetaction=hide? Again, hopefully common sense will prevail, but probably best not to rely on that.

index.html
```diff
@@ -13594,6 +13594,208 @@ <h4 id=att-playsinline>`playsinline`</h4>
...
+            If the associated element is displayed as a popover: 
+            <a class="core-mapping" href="#ariaExpandedTrue">`aria-expanded=true`</a>
+          </p>
+          <p>
+            If the associated element is hidden, or not present in the DOM: 
+            <a class="core-mapping" href="#ariaExpandedFalse">`aria-expanded=false`</a>
+          </p> 
+          <p>
+            If the associated element is an accessibility ancestor of the element with the `popovertarget` attribute:
+            <a class="core-mapping" href="#ariaExpandedUndefined">`aria-expanded=undefined`</a>
```
jcsteh, line 14321: This doesn't appear to be implemented in Chromium:
`data:text/html,<button popovertarget="pop">tog</button><div id="pop" popover>pop<button popovertarget="pop" popovertargetaction="hide">hide`
When the popover is shown, the "hide" button gets the expanded state. If I'm reading this correctly, it shouldn't.
Just to double check, is this definitely what we want and it just hasn't been done in Chromium yet? Or is this still an open question we need to sort out in this PR?

scottaohara: yes it should not be getting the expanded state for the hide button, this will need to be adjusted in chromium

index.html
```diff
@@ -14218,6 +14218,208 @@ <h4 id=att-playsinline>`playsinline`</h4>
...
+    <tr>
+      <th>[[WAI-ARIA-1.2]]</th>
+      <td>
+        <div class="general">Not mapped</div>
+      </td>
+    </tr>
+    <tr>
+      <th><a href="https://msdn.microsoft.com/en-us/library/dd373608%28v=VS.85%29.aspx">MSAA</a> + <a href="http://accessibility.linuxfoundation.org/a11yspecs/ia2/docs/html/">IAccessible2</a></th>
+      <td>
+        <span class="type">Relations:</span> `IA2_RELATION_DETAILS_FOR` points to invoking element
```
```suggestion
        <span class="type">Relations:</span> `IA2_RELATION_DETAILS_FOR` points to invoking element.  <a href="#att-popover-comments">See Comments</a>.
```

index.html
```diff
@@ -14218,6 +14218,208 @@ <h4 id=att-playsinline>`playsinline`</h4>
     </tr>
   </tbody>
 </table>
+<h4 id=att-popover>`popover`</h4>
+<table aria-labelledby=att-poster>
```
```suggestion
<table aria-labelledby=att-popover>
```

index.html
```diff
@@ -14218,6 +14218,208 @@ <h4 id=att-playsinline>`playsinline`</h4>
...
+      <th><a href="https://msdn.microsoft.com/en-us/library/ms726297%28v=VS.85%29.aspx">UIA</a></th>
+      <td>
+        ???
+      </td>
+    </tr>
+    <tr>
+      <th><a href="https://gnome.pages.gitlab.gnome.org/atk/">ATK</a></th>
+      <td>
+        <p>
+          <span class="type">Relations:</span> `RELATION_DETAILS_FOR` points to invoking element
```
```suggestion
          <span class="type">Relations:</span> `RELATION_DETAILS_FOR` points to invoking element. <a href="#att-popover-comments">See Comments</a>.
```

index.html
```diff
@@ -14218,6 +14218,208 @@ <h4 id=att-playsinline>`playsinline`</h4>
...
+    <tr>
+      <th><a href="https://developer.apple.com/reference/appkit/nsaccessibility">AX</a></th>
+      <td>
+        ???
+      </td>
+    </tr>
+    <tr>
+      <th>Comments</th>
+      <td>
+        If specified on an element with an implicit role of `generic`, then the element's role instead maps to <a class="core-mapping" href="#role-map-group">`group`</a>.
```
```suggestion
        <p>UAs MUST NOT expose a details relation between a `popover` and its invoking element under the following conditions:</p>
        <ul>
          <li>when the `popover` is the next immediate accessibility sibling to the invoking element,</li>
          <li>when the element has a `popovertarget=hide` attribute value,</li>
          <li>or when the element is a descendant of the `popover` and its `popovertarget` is set to the auto state.</li>
        </ul>
        <p>If specified on an element with an implicit role of `generic`, then the element's role instead maps to <a class="core-mapping" href="#role-map-group">`group`</a>.</p>
```

index.html
```diff
@@ -14218,6 +14218,208 @@ <h4 id=att-playsinline>`playsinline`</h4>
...
+    <tr>
+      <th>Comments</th>
+      <td>
+        <p>
+          User Agents MUST expose an <a class="core-mapping" href="#ariaDetails">`aria-details`</a> relation with the associated popover element
+          <strong>except</strong> under either of the following conditions:
+        </p>
+        <ul>
+          <li>The element's `popovertargetaction` attribute value is "`hide`"</li>
+          <li>The associated popover element is an immediate accessibility sibling to the invoking element.</li>
```
```suggestion
          <li>The associated popover element is the next immediate accessibility sibling to the invoking element,</li>
          <li>The element's `popovertargetaction` attribute is the "`auto`" state and the element is a descendant of the `popover` it is associatd with.</li>
```

index.html
```diff
@@ -14218,6 +14218,208 @@ <h4 id=att-playsinline>`playsinline`</h4>
...
+      <td>
+        <div class="general">Use WAI-ARIA mapping</div>
+      </td>
+    </tr>
+    <tr>
+      <th>Comments</th>
+      <td>
+        <p>
+          User Agents MUST expose an <a class="core-mapping" href="#ariaDetails">`aria-details`</a> relation with the associated popover element
+          <strong>except</strong> under either of the following conditions:
```
```suggestion
          <strong>except</strong> under the following conditions:
```

index.html
```diff
@@ -14218,6 +14218,215 @@ <h4 id=att-playsinline>`playsinline`</h4>
...
+        <div class="objattrs">
+          <span class="type">Object attributes:</span> 
+          <code>ispopup: &lt;value></code> where <code>&lt;value></code> reflects the <a data-cite="html/popover.html#the-popover-attribute">popover</a> type.
+        </div>
+      </td>
+    </tr>
+    <tr>
+      <th><a href="https://developer.apple.com/reference/appkit/nsaccessibility">AX</a></th>
+      <td>
+        ???
```
```suggestion
        Not mapped
        <!-- https://bugs.webkit.org/show_bug.cgi?id=271485 -->
```

index.html
```diff
@@ -14218,6 +14218,215 @@ <h4 id=att-playsinline>`playsinline`</h4>
...
+        <div class="objattrs">
+          <span class="type">Object attributes:</span>
+          <code>ispopup: &lt;value></code> where <code>&lt;value></code> reflects the <a data-cite="html/popover.html#the-popover-attribute">popover</a> type.
+        </div>
+      </td>
+    </tr>
+    <tr>
+      <th><a href="https://msdn.microsoft.com/en-us/library/ms726297%28v=VS.85%29.aspx">UIA</a></th>
+      <td>
+        ???
```
```suggestion
        A details relation is made with the invoking element, if an invoking element exists which meets the conditions for necessitating a details relationship. <a href="#att-popover-comments">See Comments</a>.
```

index.html
benbeaudry: Have we filed a bug for this?

index.html
```diff
@@ -14218,6 +14218,216 @@ <h4 id=att-playsinline>`playsinline`</h4>
...
+      <th>Comments</th>
+      <td>
+        <p>
+          User Agents MUST expose an <a class="core-mapping" href="#ariaDetails">`aria-details`</a> relation with the associated popover element
+          <strong>except</strong> under the following conditions:
+        </p>
+        <ul>
+          <li>The element's `popovertargetaction` attribute value is "`hide`"</li>
+          <li>The associated popover element is the next immediate accessibility sibling to the invoking element,</li>
+          <li>The element's `popovertargetaction` attribute is the "`auto`" state and the element is a descendant of the `popover` it is associatd with.</li>
```
benbeaudry: nit: typo in "associatd"

benbeaudry (approved):
LGTM

index.html
```suggestion
          <li>The element's `popovertargetaction` attribute is the "`auto`" state and the element is a descendant of the `popover` it is associated with.</li>
```

scottaohara:
closing - will be merged via https://github.com/w3c/aria/pull/2219

PR #484, opened 2023-05-11
contextual role concept proposal

### LINKED ISSUE #375: Define "scoped" or reuse another existing term?
This is follow on work to #350 as HTML AAM (and ARIA in HTML) use the term 'scoped' to indicate when a `header`, `footer` and `aside` have an implicit ARIA landmark role, or not.   "Scoped" in this context essentially means that they are an ancestor of the `body` element (or the `body` or `main` element, in `aside`'s case), with no intervening sectioning content or sectioning root elements.

There are a few things that need to be addressed with this term:

1. it is not formally defined as a term at this point.  This can be done, but as there are similar concepts in ARIA (required context role / Owned Element) and in HTML (Contexts in which this element can be used), figured it would make sense to get the WG's opinion on if a term name change might be in order.  I don't presently think we can just use one of these other terms... but open to being told otherwise - there might be overlap with https://github.com/w3c/aria/pull/1454
2. The current wording for scoped elements doesn't take into account what should happen if someone were to specify an intervening element (`div`) - which is fine for the scoping rule, but gives it an explicit ARIA role which effectively exposes it as an element which _should_ break the scoping rule. 
3. I think this concept is important to specify beyond its current use for `header`, `footer` and `aside`.  There are other elements that have similar behavior of becoming 'generic' or having their role 'ignored' under certain circumstances. But as far as i can tell, this isn't actually specified so much as it just is accepted behavior.  e.g., a list item is not exposed as a list item if not contained within a list.  Or if a table's implicit role is overwritten, all its direct table descendants become generics... and if you say use a `td` outside of the context of a table, browsers don't even render the element - its contents are just rendered instead.

Interested in people's thoughts on this.

scottaohara:
closes #375

The purpose of this proposal is both to resolve the issue that "scoped" was an undefined term that was used to indicate when certain elements would expose particular ARIA roles or not.  However, in the year of noodling on this, other activities to revise what "owned" means in ARIA, the idea of a minimum role, as well as the fact there were other elements in HTML that have "scoped" or "contextual roles" that are based on heuristics, rather than defined standards, became relevant.

A change to hopefully help simplify the implementations of contextual roles is that this proposal is aiming to tie the role of an element based on its a11y tree ancestry, rather than based on specific HTML elements in the DOM.

This PR represents the initial draft of the proposal.  Other work to explicitly call out the elements that have contextual roles, what roles they need to adjust to, and other wording revisions to replace "scoped" with the new term would still need to be made.

## Implementation

* WPT tests: [LINK]()
   * WebKit: [ISSUE]()
   * Gecko: [ISSUE]()
   * Blink: [ISSUE]()

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/html-aam/pull/484.html" title="Last updated on Nov 20, 2023, 7:00 PM UTC (4f13392)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/html-aam/484/14cdf18...4f13392.html" title="Last updated on Nov 20, 2023, 7:00 PM UTC (4f13392)">Diff</a>


index.html
```diff
@@ -210,6 +210,71 @@ <h3>Exposing HTML Features That Do Not Directly Map to Accessibility APIs</h3>
         <abbr title="application programming interface">API</abbr> properties</a> in the [[core-aam-1.2]].
       </p>
     </section>
+    <section id="mapping_contextual">
+      <h3>Exposing HTML Features With Contextual Roles</h3>
+      <p>
+        An element with a <dfn>contextual role</dfn> is one whose implicit WAI-ARIA role semantics are based on its anestral <a>accessibility tree</a> relationship to
```
```suggestion
        An element with a <STRONG>contextual role</STRONG> is one whose implicit WAI-ARIA role semantics are based on its ancestral <a>accessibility tree</a> relationship to 
```

index.html
```diff
@@ -210,6 +210,71 @@ <h3>Exposing HTML Features That Do Not Directly Map to Accessibility APIs</h3>
...
+        An element with a <dfn>contextual role</dfn> is one whose implicit WAI-ARIA role semantics are based on its anestral <a>accessibility tree</a> relationship to 
+        another element or elements. Or, an element whose computed role is reliant on whether the element has been provided an accessible name. 
+      </p>
+      <p>
+        For instance, an element that is an <STRONG>accessibility child</STRONG> of its required <STRONG>accessibility parent</STRONG> would expose a computed role that was 
+        appropriate for its related context. Where the same element with a contextual role might be used within another markup pattern and have a different accessibility parent, 
+        and would therefore expose a computed role appropriate for that context.
+      </p>
+      <p>
+        When the conditions for an element's contextual role are not met, it will commonly expose a computed role of `generic` or with a <a>minimum role</a>, if applicable.
```
```suggestion
        When the conditions for an element's contextual role are not met, it will commonly expose a computed role of `generic` or with a <STRONG>minimum role</STRONG>, if applicable. 
```

index.html
```diff
@@ -210,6 +210,71 @@ <h3>Exposing HTML Features That Do Not Directly Map to Accessibility APIs</h3>
...
+        &lt;li>...&lt;/li> &lt;!-- role=listitem -->
+      &lt;/ul> 
+      &lt;article>
+        ...
+        &lt;li>...&lt;/li> &lt;!-- not role=listitem -->
+        ...</code></pre>
+      </div>
+      <div class="example">
+        <p>
+          An <code>aside</code> element has an implicit WAI-ARIA role mapping of <code>complementary</code> only when the element is an <a>accessibility child</a> of
```
```suggestion
          An <code>aside</code> element has an implicit WAI-ARIA role mapping of <code>complementary</code> only when the element is an <STRONG>accessibility child</STRONG> of 
```

index.html
```diff
@@ -210,6 +210,71 @@ <h3>Exposing HTML Features That Do Not Directly Map to Accessibility APIs</h3>
...
+        &lt;aside>...&lt;/aside> &lt;!-- role=complementary -->
+        ...
+        &lt;footer>
+          ...
+          &lt;aside>...&lt;/aside> &lt;!-- not role=complementary -->
+        &lt;/footer></code></pre>
+      </div>
+
+      <div class="example">
+        <p>A <code>section</code> element has an implicit WAI-ARIA role mapping of <code>region</code> only when the element is provided an <a>accessible name</a>.</p>
```
```suggestion
        <p>A <code>section</code> element has an implicit WAI-ARIA role mapping of <code>region</code> only when the element is provided an <STRONG>accessible name</STRONG>.</p>
```

scottaohara (commented):
following quick commits are additional instances of converting undefined terms to temporary STRONG elements.  fix of a typo

index.html
```diff
@@ -210,6 +210,71 @@ <h3>Exposing HTML Features That Do Not Directly Map to Accessibility APIs</h3>
         <abbr title="application programming interface">API</abbr> properties</a> in the [[core-aam-1.2]].
       </p>
     </section>
+    <section id="mapping_contextual">
+      <h3>Exposing HTML Features With Contextual Roles</h3>
+      <p>
+        An element with a <STRONG>contextual role</STRONG> is one whose implicit WAI-ARIA role semantics are based on its ancestral <a>accessibility tree</a> relationship to 
+        another element or elements. Or, an element whose computed role is reliant on whether the element has been provided an accessible name.
```
spectranaut: Do we want to start off with something even vaguer and more all encompassing, for example:

> An element with a contextual role is one whose implicit WAI-ARIA role semantics are dependent on the existence of other elements in the accessibility tree or the existence of certain attributes. 

Maybe that is too vague? Maybe is an "accessible name" an "attribute"? Can we imagine other attributes might effect a role? Like `tabindex`?

Regardless I do object to a sentence starting with `Or, `. Here is another try, just to try things out:

> An element with a contextual role is one whose implicit WAI-ARIA role semantics are effected by one or more of the following constraints:
> - A required ancestral relationship with other elements in the accessibility tree.
> - The existence of an accessible name.

spectranaut (changes requested):
I think it would be nice to see at lease one example of the new "contextual role" term being used in the mapping tables.

I really think this is the right direction!! I can help nit pick the specific wording ad infinitum though.

index.html
spectranaut: oh wow when I mentioned tabindex, I just remembered this had an effect on the "minimum role" concept. Honestly, it seems, minimum role is actually a "contextual role", and maybe... we don't need these things to be separately defined?

scottaohara: i did have that thought as well... but the rational i kept coming back to as to why they should be separate is that contextual roles are elements that have defined roles, but do to both valid and invalid use, their role may need to adjust.  These behaviors largely exist now as either defined for specific elements, or are undefined behaviors that presently occur.

where as minimum roles are really only for elements that aren't meant to be anything more than generic or not exposed to the a11y api, but due to valid use of HTML, they need to have a minimum role based on the attribute that was used on the element.  minimum roles are not presently implemented, unlike most if not all of the contextual roles.

the distinctions could be clearer with further edits... but i do completely agree that the concepts are related and could be merged into a single section.  Just two different tracks, where i really don't want one concept to be held up by the other (minimum role being more important imo)

spectranaut: ok I can easily see them as both separate or under the same umbrella, and I see your point about not holding things up, I'm fine with keeping them separate.

cookiecrook: I like the first sentence phrasing of "ancestral" context, but I have a suggestion for the second to broaden it, but keep the name/label as an concise, understandable example.

```suggestion
        another element or elements. Or, an element whose computed role is dependent upon whether it has other associated accessibility information, such as an author-provided accessible label. 
```

index.html
```diff
@@ -210,6 +210,71 @@ <h3>Exposing HTML Features That Do Not Directly Map to Accessibility APIs</h3>
...
+    <section id="mapping_contextual">
+      <h3>Exposing HTML Features With Contextual Roles</h3>
+      <p>
+        An element with a <STRONG>contextual role</STRONG> is one whose implicit WAI-ARIA role semantics are based on its ancestral <a>accessibility tree</a> relationship to 
+        another element or elements. Or, an element whose computed role is reliant on whether the element has been provided an accessible name. 
+      </p>
+      <p>
+        For instance, an element that is an <STRONG>accessibility child</STRONG> of its required <STRONG>accessibility parent</STRONG> would expose a computed role that was 
+        appropriate for its related context. Where the same element with a contextual role might be used within another markup pattern and have a different accessibility parent, 
+        and would therefore expose a computed role appropriate for that context.
```
cookiecrook: Word-smithing plus a few simple inline examples.

```suggestion
        For instance, an element that is an <STRONG>accessibility child</STRONG> of its required <STRONG>accessibility parent</STRONG> would expose a computed role for its given context, such as a list item (<code>li</code>) within an unordered list (<code>ul</code>). However, the same element might be used within a different markup pattern and could therefore expose different contextually-dependent computed roles, such as a labeled <code>section</code> element (role <code>region</code>) versus an unlabeled <code>section</code> element (role <code>generic</code>).
```

Can we think of a specific example that works for both? If so that'd be better than the sectioning content examples, which are less easy to understand and care about than a simple list.

index.html
```diff
@@ -210,6 +210,71 @@ <h3>Exposing HTML Features That Do Not Directly Map to Accessibility APIs</h3>
...
+      </p>
+      <p>
+        When the conditions for an element's contextual role are not met, it will commonly expose a computed role of `generic` or with a <STRONG>minimum role</STRONG>, if applicable.
```
cookiecrook, line 225: consistency nit
```suggestion
        When the conditions for an element's contextual role are not met, it will commonly expose a computed role of `generic` or with a <strong>minimum role</strong>, if applicable. 
```

index.html
```diff
@@ -210,6 +210,71 @@ <h3>Exposing HTML Features That Do Not Directly Map to Accessibility APIs</h3>
...
+      </p>
+      <p>
+        When the conditions for an element's contextual role are not met, it will commonly expose a computed role of `generic` or with a <STRONG>minimum role</STRONG>, if applicable. 
+      </p>
+      <div class="note">
+        <p>Previously, the concept of a contextual role was loosely referred to as an element's "scoped" relationship to another element.</p>
+      </div>
+      <div class="example">
+        <p>
+          A <code>li</code> element has an implicit WAI-ARIA role mapping of <code>listitem</code> only when the element is an <STRONG>accessibility child</STRONG> of a
```
cookiecrook, line 232:
```suggestion
          A <code>li</code> element has an implicit WAI-ARIA role mapping of <code>listitem</code> only when the element is an <strong>accessibility child</strong> of a 
```

index.html
```diff
@@ -210,6 +210,71 @@ <h3>Exposing HTML Features That Do Not Directly Map to Accessibility APIs</h3>
...
+      <div class="note">
+        <p>Previously, the concept of a contextual role was loosely referred to as an element's "scoped" relationship to another element.</p>
+      </div>
+      <div class="example">
+        <p>
+          A <code>li</code> element has an implicit WAI-ARIA role mapping of <code>listitem</code> only when the element is an <STRONG>accessibility child</STRONG> of a 
+          <code>menu</code>, <code>ol</code> or <code>ul</code> element that has not had its `list` role mapping suppressed.
+        </p>
+        <pre><code>&lt;ul>
+        &lt;li>...&lt;/li> &lt;!-- role=listitem -->
```
cookiecrook:
```suggestion
        &lt;li>...&lt;/li> &lt;!-- computed role `listitem` -->
```

index.html
```diff
@@ -210,6 +210,71 @@ <h3>Exposing HTML Features That Do Not Directly Map to Accessibility APIs</h3>
...
+      &lt;/ul> 
+      &lt;article>
+        ...
+        &lt;li>...&lt;/li> &lt;!-- not role=listitem -->
```
cookiecrook:
```suggestion
        &lt;li>...&lt;/li> &lt;!-- computed role `generic`, not `listitem` -->
```

index.html
```diff
@@ -210,6 +210,71 @@ <h3>Exposing HTML Features That Do Not Directly Map to Accessibility APIs</h3>
...
+        ...</code></pre>
+      </div>
+      <div class="example">
+        <p>
+          An <code>aside</code> element has an implicit WAI-ARIA role mapping of <code>complementary</code> only when the element is an <STRONG>accessibility child</STRONG> of
```
cookiecrook, line 245:
```suggestion
          An <code>aside</code> element has an implicit WAI-ARIA role mapping of <code>complementary</code> only when the element is an <strong>accessibility child</strong> of 
```

index.html
```diff
@@ -210,6 +210,71 @@ <h3>Exposing HTML Features That Do Not Directly Map to Accessibility APIs</h3>
...
+            &lt;aside aria-label=named>...&lt;/aside> &lt;!-- role=complementary -->
+          &lt;/article>
+          ...
+          &lt;aside>...&lt;/aside> &lt;!-- role=complementary -->
+        &lt;/main>
+        &lt;aside>...&lt;/aside> &lt;!-- role=complementary -->
+        ...
+        &lt;footer>
+          ...
+          &lt;aside>...&lt;/aside> &lt;!-- not role=complementary -->
```
cookiecrook:
```suggestion
            &lt;aside>...&lt;/aside> &lt;!-- computed role `generic`, not `complementary` -->
            ...
            &lt;aside aria-label=named>...&lt;/aside> &lt;!-- computed role `complementary` -->
          &lt;/article>
          ...
          &lt;aside>...&lt;/aside> &lt;!-- computed role `complementary` -->
        &lt;/main>
        &lt;aside>...&lt;/aside> &lt;!-- computed role `complementary` -->
        ...
        &lt;footer>
          ...
          &lt;aside>...&lt;/aside> &lt;!-- computed role `generic`, not `complementary` -->
```

index.html
```diff
@@ -210,6 +210,71 @@ <h3>Exposing HTML Features That Do Not Directly Map to Accessibility APIs</h3>
...
+        &lt;/footer></code></pre>
+      </div>
+
+      <div class="example">
+        <p>A <code>section</code> element has an implicit WAI-ARIA role mapping of <code>region</code> only when the element is provided an <STRONG>accessible name</STRONG>.</p>
```
cookiecrook, line 269:
```suggestion
        <p>A <code>section</code> element has an implicit WAI-ARIA role mapping of <code>region</code> only when the element is provided an <strong>accessible name</strong>.</p>
```

index.html
```diff
@@ -210,6 +210,71 @@ <h3>Exposing HTML Features That Do Not Directly Map to Accessibility APIs</h3>
...
+
+      <pre><code>&lt;section title="accessible name">...&lt;/section> &lt;!-- role=region -->
+      ...
+      &lt;section aria-label="accessible name">...&lt;/section> &lt;!-- role=region -->
+      ...
+      &lt;section>...&lt;/section> &lt;!-- not role=region --></code></pre>
```
cookiecrook:
```suggestion
      <pre><code>&lt;section title="accessible name">...&lt;/section> &lt;!-- computed role `region` -->
      ...
      &lt;section aria-label="accessible name">...&lt;/section> &lt;!-- computed role `region` -->
      ...
      &lt;section>...&lt;/section> &lt;!-- computed role `generic`, not `region` --></code></pre>
```

cookiecrook (changes requested):

index.html
scottaohara: the STRONGs are only in there as placeholders, since I would need to remove these to reference the terms which haven't been published yet, so were breaking the HTML AAM build.   i should have been more clear on that when intro-ing the draft.

cookiecrook: okay. I'll delete the others and adjust the previous one to not include that edit.

index.html
spectranaut: nice I like this suggestion, @cookiecrook

index.html
scottaohara: @cookiecrook I like these suggestions so i'm going to largely pull them in - but I'm going to be changing the use of "accessible label" back to "name".

scottaohara: @cookiecrook re: other examples.

What do you think about mentioning tbody here, per the whole rowgroup conversation?  That'd be an example of a contextual role where it can be ignored until it is given a name or focusable.

Another example we could use in lieu of / in combination with the list item example is the summary of a details, where only under specific conditions is the summary an `html-summary`, and otherwise it would need to be a `role=generic`

cookiecrook: 🤷 I could go either way. I don't think most authors give much thought to `<tbody>` etc, so I'm not sure it'd be super helpful as an example, but I also don't think it'd be harmful.

index.html
```diff
@@ -210,6 +210,71 @@ <h3>Exposing HTML Features That Do Not Directly Map to Accessibility APIs</h3>
         <abbr title="application programming interface">API</abbr> properties</a> in the [[core-aam-1.2]].
       </p>
     </section>
+    <section id="mapping_contextual">
+      <h3>Exposing HTML Features With Contextual Roles</h3>
```
rahimabdi: This heading may not accurately capture the intent of the section; particularly, the words "HTML Features" used to describe contextual roles and how they're treated. I don't know the best way to describe the new section but if you agree, some alternatives for your consideration: Contextual Roles, Treatment of Contextual Roles, Understanding Contextual Roles, Understanding Contextual Roles and Accessibility Ancestry

rahimabdi:
- https://github.com/w3c/aria/pull/2218

PR #497, opened 2023-07-25
Clarify accName for input type=image

### LINKED ISSUE #414: Input type="image" - Accessible name computation
Hi everyone,

the ACT Rules Community Group is improving an atomic rule to ensure image button has non-empty accessible name (https://deploy-preview-103--wai-wcag-act-rules.netlify.app/standards-guidelines/act/rules/59796f/).

We're discussing this topic in:
- https://github.com/act-rules/act-rules.github.io/issues/1861
- https://github.com/act-rules/act-rules.github.io/pull/1865

We're a bit concerned about how technologies (both browsers and AT) render/expose inconsistently elements that support the alt attribute, in particular input type="image" (but also img and area), especially with particular combination of attributes set (e.g. empty alt and non-empty title, empty aria-label and non-empty title, etc.).

In addition, following the [4.3.2 input type="image" Accessible Description Computation](https://www.w3.org/TR/html-aam-1.0/#input-type-image-accessible-name-computation), from steps 2  the alt attribute should be used as accessible name if present, from step 3 the title attribute should be used as accessible name if present. However, step 4  states "if the previous steps do not yield a usable text string" where it's not clear if an empty string should count as a "usable text string" or not.

We would have your opinion on what the "good" way should be to write our rule accordingly.

We've also created a html file containing the entire list of examples for every element that supports alt (img, input type="image" and area) with the combination of title and other properties (title + empty alt, title + empty aria-label and title + empty aria-labelledby). We've added valid and invalid images due to different behaviour with Firefox:
- empty alt (alt="") and non-empty title is behaving consistently both with loaded and non-loaded image
- empty aria-label (aria-label="") and empty aria-labelledby (aria-labelledby="") are behaving inconsistently

Please, let us know your thoughts.

Thanks,
Giacomo

[img-inputimg-area-examples.zip](https://github.com/w3c/html-aam/files/8930974/img-inputimg-area-examples.zip)

scottaohara:
adds clarifying language for the alt and title steps.  

updates the localized text string step to better match the updated input type=button/reset/submit, and note the fact that this element is considered a 'submit button' - linking to that portion of the HTML spec and noting that different browsers may provide a different default string (e.g., chrome does 'submit' while firefox does 'submit query')

Closes #414

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/html-aam/pull/497.html" title="Last updated on Jul 25, 2023, 1:45 PM UTC (075f7c4)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/html-aam/497/551a9e9...075f7c4.html" title="Last updated on Jul 25, 2023, 1:45 PM UTC (075f7c4)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -6175,12 +6175,14 @@
           <li>
             Otherwise use the associated `label` element(s) <a data-cite="accname-1.2/#dfn-accessible-name">accessible name</a>(s) - if more than one `label` is associated; concatenate by DOM order, delimited by spaces.
           </li>
-          <li>Otherwise use `alt` attribute.</li>
-          <!-- NOTE: use of valid attribute is invalid on input type=image, but it DOES contribute to the name if used -->
+          <li>Otherwise use `alt` attribute if present and its value is not the empty string.</li>
+          <!-- NOTE: use of value attribute is invalid on input type=image, but it DOES contribute to the name if used -->
           <!-- <li>Otherwise use `value` attribute.</li> -->
-          <li>Otherwise use `title` attribute.</li>
+          <li>Otherwise use `title` attribute if present and its value is not the empty string.</li>
           <li>
-            Otherwise if the previous steps do not yield a usable text string, the <a data-cite="accname-1.2/#dfn-accessible-name">accessible name</a> is a localized string of the word &quot;Submit Query&quot;.
+            Otherwise if the previous steps do not yield a usable text string, use the 
+            <a href="https://infra.spec.whatwg.org/#implementation-defined">implementation defined</a> string respective to the input type (an `input` in the `image` state
+            represents a <a data-cite="html/forms.html#concept-submit-button">submit button</a>). For instance, a localized string of the word "submit" or the words &quot;Submit Query&quot;.
           </li>
           <li>
             If none of the above yield a usable text string there is no <a data-cite="accname-1.2/#dfn-accessible-name">accessible name</a>.


index.html
```diff
@@ -6175,12 +6175,14 @@ <h4>`input type="image"` Accessible Name Computation</h4>
           <li>
             Otherwise use the associated `label` element(s) <a data-cite="accname-1.2/#dfn-accessible-name">accessible name</a>(s) - if more than one `label` is associated; concatenate by DOM order, delimited by spaces.
           </li>
-          <li>Otherwise use `alt` attribute.</li>
-          <!-- NOTE: use of valid attribute is invalid on input type=image, but it DOES contribute to the name if used -->
+          <li>Otherwise use `alt` attribute if present and its value is not the empty string.</li>
```
MelSumner, line 6178:
```suggestion
          <li>Otherwise use `alt` attribute if present and its value is not an empty string.</li>
```

index.html
```diff
@@ -6175,12 +6175,14 @@ <h4>`input type="image"` Accessible Name Computation</h4>
...
+          <!-- NOTE: use of value attribute is invalid on input type=image, but it DOES contribute to the name if used -->
           <!-- <li>Otherwise use `value` attribute.</li> -->
-          <li>Otherwise use `title` attribute.</li>
+          <li>Otherwise use `title` attribute if present and its value is not the empty string.</li>
```
MelSumner, line 6181:
```suggestion
          <li>Otherwise use `title` attribute if present and its value is not an empty string.</li>
```

MelSumner (approved):
Suggestion, non-blocker: I think "an empty string" reads a little better than "the empty string". 👍

accdc (approved):

cookiecrook (approved):

index.html
scottaohara: Don’t disagree, but I’m not going to break convention with how this concept is consistently referred to in specs.

PR #498, opened 2023-08-16
add mention of image role

scottaohara:
Firefox and Safari have implemented/released the update to make image a synonym role for img.  Chromium will be releasing this update soon.

this PR references the image element as mapping to img/image, and adds a comment to reference the synonym addition in ARIA 1.3

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/html-aam/pull/498.html" title="Last updated on Oct 3, 2023, 2:22 PM UTC (2b7409c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/html-aam/498/06c41c2...2b7409c.html" title="Last updated on Oct 3, 2023, 2:22 PM UTC (2b7409c)">Diff</a>


cookiecrook:
Also note that @spectranaut added this to the computed role column, so it's in the spec already. Fine to add another note though.

cookiecrook:
> Firefox and Safari have implemented/released the update to make image a synonym role for img. Chromium will be releasing this update soon.

And to clarify, this means `img` is a synonym of the newer-but-now-primary role `image` role. Likewise the original `presentation` is now a synonym of the preferred/primary 1.1 role name, `none`…

index.html
```diff
@@ -1389,14 +1389,24 @@ <h3>HTML Element Role Mappings</h3>
               <th>
                 <a data-cite="HTML">`img`</a>
               </th>
-              <td class="aria"><a class="core-mapping" href="#role-map-img">`img`</a> role</td>
+              <td class="aria">
+                <a class="core-mapping" href="#role-map-img">`img`</a> role
+                or
+                <!-- todo: update role map link if/when core aam adds in a specific image role mapping table -->
```
scottaohara: i question the value in having separate tables in core aam for presentation|none and potentially now img|image

could these not just be one table?

index.html
```diff
@@ -1389,14 +1389,24 @@ <h3>HTML Element Role Mappings</h3>
...
               <td class="role-computed"><div class="general">Use WAI-ARIA mapping</div></td>
               <td class="ia2"><div class="general">Use WAI-ARIA mapping</div></td>
               <td class="uia"><div class="general">Use WAI-ARIA mapping</div></td>
               <td class="atk"><div class="general">Use WAI-ARIA mapping</div></td>
               <td class="ax"><div class="general">Use WAI-ARIA mapping</div></td>
               <!-- <td class="naming"></td> -->
-              <td class="comments"></td>
+              <td class="comments">
+                <div class="note">
+                  ARIA 1.3 adds the `image` role as a synonym of the ARIA 1.0 `img` role.
```
scottaohara: note i know that image is the primary role - but this is using text directly from a note in the ARIA spec.  if we want to change this wording, then we should change the ARIA spec wording too

index.html
spectranaut: Hmm I see your point, but I don't think its necessary. I think it's ok to be verbose and repeat here, these specs are edited rarely, and I think it makes sense to have an entry for every role. Then we will have a unique link to every role (#role-map-img and #role-map-image). To combine them would introduce a different kind of complexity.

index.html
spectranaut: I think this wording is fine.

index.html
```diff
@@ -1389,14 +1389,24 @@ <h3>HTML Element Role Mappings</h3>
               <th>
                 <a data-cite="HTML">`img`</a>
               </th>
-              <td class="aria"><a class="core-mapping" href="#role-map-img">`img`</a> role</td>
+              <td class="aria">
+                <a class="core-mapping" href="#role-map-img">`img`</a> role
+                or
+                <!-- todo: update role map link if/when core aam adds in a specific image role mapping table -->
+                <a class="core-mapping" href="#role-map-img">`image`</a>
```
```suggestion
                <a class="core-mapping" href="#role-map-image">`image`</a>
```

As I said before, I do think there should be (as there is now) an entry for every role in CORE-AAM. And this link should go to "image" role as it suggests.

spectranaut (changes requested):

index.html
scottaohara: then this PR is blocked on core aam not having an image role entry and adding that is semi-blocked on changing core aam to move away from the disclosure widgets, unless we want to do this change and then merge it into that larger work.

cc @spectranaut

scottaohara: or we can merge as is, and then update the link once the image role is added to core aam.

i prefer doing that for something so minor, like this.  but will wait if others disagree

spectranaut: omg! I was confused by your comment because I was so convinced there was an entry for "image". Tomorrow I'm going to merge the changes related to removing the tables/disclosure, I'll also open a PR adding image, and then you can merge this, how about that?

spectranaut (approved):

index.html
```diff
@@ -1389,14 +1389,24 @@ <h3>HTML Element Role Mappings</h3>
               <th>
                 <a data-cite="HTML">`img`</a>
               </th>
-              <td class="aria"><a class="core-mapping" href="#role-map-img">`img`</a> role</td>
+              <td class="aria">
+                <a class="core-mapping" href="#role-map-img">`img`</a> role
```
cookiecrook: Since `image` is the primary, shouldn't this just be "`image`", not "`img` **_or_** `image`"

index.html
```diff
@@ -1407,6 +1417,8 @@ <h3>HTML Element Role Mappings</h3>
               <td class="aria">
                 <div class="role">
                   <a class="core-mapping" href="#role-map-presentation">`presentation`</a>
+                  or
+                  <a class="core-mapping" href="#role-map-none">`none`</a>
```
cookiecrook: Same issue here... IMO. should be "`none`", not "`presentation` or `none`"

cookiecrook (changes requested):

scottaohara:
have fixed the merge conflicts, and have addressed @cookiecrook's comments.  I"m going to merge this PR and can fast follow with any additional changes / create new issues/prs as necessary

PR #522, opened 2023-12-04
feat: add IA2 and ATK mappings for text edit maxlength

smhigley:
Part of the initial work for https://github.com/w3c/aria/issues/1119

Splitting out each API into a separate PR for ease of review and discussion. I've combined IA2 and ATK here since they generally seem to have the same approach re: Object Attributes, and existing object attributes look like they have the same keys across both APIs

***
<a href="https://pr-preview.s3.amazonaws.com/smhigley/html-aam/pull/522.html" title="Last updated on Dec 13, 2023, 11:11 PM UTC (208726b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/html-aam/522/ab38b66...smhigley:208726b.html" title="Last updated on Dec 13, 2023, 11:11 PM UTC (208726b)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -12282,7 +12282,9 @@
     <tr>
       <th><a href="https://gnome.pages.gitlab.gnome.org/atk/">ATK</a></th>
       <td>
-        <div class="general">Not mapped</div>
+        <div class="objattrs">
+          <span class="type">Object attributes:</span> `max-edit-chars: &lt;value&gt;`
+        </div>
       </td>
     </tr>
     <tr>
--- a/index.html
+++ b/index.html
@@ -12269,7 +12269,7 @@
       <th><a href="https://msdn.microsoft.com/en-us/library/dd373608%28v=VS.85%29.aspx">MSAA</a> + <a href="http://accessibility.linuxfoundation.org/a11yspecs/ia2/docs/html/">IAccessible2</a></th>
       <td>
         <div class="objattrs">
-          <span class="type">Object attributes:</span> `max-edit-chars: &lt;value&gt;`
+          <span class="type">Object attributes:</span> `maxlength: &lt;value&gt;`
         </div>
       </td>
     </tr>
@@ -12283,7 +12283,7 @@
       <th><a href="https://gnome.pages.gitlab.gnome.org/atk/">ATK</a></th>
       <td>
         <div class="objattrs">
-          <span class="type">Object attributes:</span> `max-edit-chars: &lt;value&gt;`
+          <span class="type">Object attributes:</span> `maxlength: &lt;value&gt;`
         </div>
       </td>
     </tr>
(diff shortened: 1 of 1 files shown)


joanmarie:
Is there going to be a corresponding ARIA property now or in the future? If so, will it be `aria-max-edit-chars`?

In general, when ARIA properties are exposed as object attributes, the rule tends to be `aria-foo` is exposed as `foo`. So if the answer to my second question is "yes", then this change looks good to me. But if the ARIA property will have a different name, please figure out what that is first and update this mapping accordingly.

Thanks for checking!

smhigley:
@joanmarie I believe the planned aria attr would be `aria-maxlength`, should the object attribute just be `maxlength` then?

`max-edit-chars` was one of the ones @aleventhal suggested, but that was just a quick chat and not a super in-depth dive. Also if you want this to be a draft until the ARIA attribute is in PR, or go directly to Core AAM, let me know. I just didn't want to hold up browser & screen reader work till the ARIA attribute is added.

joanmarie:
> @joanmarie I believe the planned aria attr would be `aria-maxlength`, should the object attribute just be `maxlength` then?

I think so, yes -- unless @aleventhal has a compelling reason why it should be exposed to screen readers as `max-edit-chars`. As the Orca maintainer, I cannot think of one. But I don't know if there are concerns on the Windows side of things.

I also think this will make it possible for you to get at least one ARIA implementation "for free" (i.e. Gecko). I just loaded the following in Firefox:

`data:text/html,<div role="textbox" aria-maxlength="50" aria-doesnotexist="oh really?">foo</div>`

Then I used the Accerciser accessibility inspector in Linux to query the properties of the AtspiObject that corresponds to the div:

* `acc.getRole()` returns `<enum ATSPI_ROLE_ENTRY of type Atspi.Role>`
* `acc.queryText().getText(0,-1)` returns `'foo'`
* `acc.getAttributes()` returns `['display:block', 'maxlength:50', 'tag:div', 'doesnotexist:oh really?', 'xml-roles:textbox']`

Gecko doesn't sanity-check or rename ARIA attributes to object attributes for ATK or IA2 -- other than stripping off the "aria" prefix -- unless there's a strong reason to do so.  :tada: :grin: 

>  I just didn't want to hold up browser & screen reader work till the ARIA attribute is added.

Understood. And I don't want to hold up spec work. Perhaps check with @jcsteh at Mozilla regarding his preference and thoughts about `max-edit-chars` being the attribute to be exposed for HTML?

jcsteh:
I'd prefer the object attribute to be named according to the ARIA attribute. For example, if it's aria-maxlength, I think the object attribute should be maxlength. Beyond that, I don't have a strong preference what the ARIA attribute is called.

smhigley:
@joanmarie / @jcsteh updated to `maxlength`!

joanmarie (approved):
LGTM. Thanks! I'll add this to my Orca TODO list.

jcsteh:
I assume that when we get aria-maxlength, this mapping will be updated so that it simply refers to the ARIA mapping and core-aam will absorb the platform specific guidance here?

scottaohara:
@jcsteh yes, that would be the end goal.  if an aria attribute is created, then the mappings go into core aam, and html aam is updated to point to that.

scottaohara:
closing as this PR has been superseded by https://github.com/w3c/aria/pull/2273

smhigley:
For reference, this is the UIA PR: https://github.com/w3c/aria/pull/2272

PR #530, opened 2024-02-06
Update: change UIA control type for label element

### LINKED ISSUE #524: Update UIA mappings for the <label> element
The current mappings for the <label> element indicate that a <label> element should be mapped to a UIA_TextControlTypeId. However, since a <label> element can have children, it's possible to have cases like this: <label>my label</label> which would then be exposed to UIA as 

++UIA_TextControlTypeId
++++UIA_TextControlTypeId

This is invalid in the world of UIA. UIA expects UIA_TextControlTypeId to be a leaf node. It effectively makes Narrator, a UIA-based AT, read both UIA_TextControlTypeId nodes independently. Instead, I suggest we map the <label> element to UIA_GroupControlTypeId and continue mapping its descendants like they should be. We just performed that change in Chromium, and it leads to an improved user experience with UIA based ATs.

I'm wondering if this could possibly lead to undefined behaviors, though. Thoughts?

scottaohara:
closes #524

this may be the first of many changes that need to be made in HTML AAM and Core AAM, based on the rational of why this control type needed to be updated.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/html-aam/pull/530.html" title="Last updated on Feb 26, 2024, 8:13 PM UTC (796f79d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/html-aam/530/3a5bd10...796f79d.html" title="Last updated on Feb 26, 2024, 8:13 PM UTC (796f79d)">Diff</a>


benbeaudry (approved):
LGTM. Thanks!

spectranaut (approved):

PR #533, opened 2024-02-26
Clarify hidden elements do not return name

### LINKED ISSUE #370: Consider: hidden <label> can still name associated form control
See also: https://github.com/w3c/accname/issues/148

The intent here is to clarify in HTML AAM that a hidden `<label>` element which is associated with a form control should still take part in the accName calculation

e.g.,
```
<label hidden for=f>foo</label>
<input id=f>
```
the input should still have a name of "foo"

scottaohara:
closes #370

Adds comments to label, legend and caption elements to indicate that if they are hidden, then they will not return a name for their associated elements.

Note that right now this means that 
```
<label for=d hidden>foo</label>
<input id=d title=bar>
```
has _no name_.  The label is still referenced, but because it is hidden it returns an empty name in Safari and Chromium browsers.  Firefox correctly returns "bar" as the name of the input.

I think this can probably be made more clear in the naming steps, and I can make some wpts for this... i'm just not sure exactly where to put them now without potentially overlapping or ballooning out some of the existing tests.  probably worth talking about that when we triage this.

The PR also includes some consistency cleanup for legend/caption.  

## Implementation

* WPT tests:
   * WebKit:
   * Gecko:
   * Blink:

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/html-aam/pull/533.html" title="Last updated on Mar 7, 2024, 1:28 PM UTC (45695fb)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/html-aam/533/e644368...45695fb.html" title="Last updated on Mar 7, 2024, 1:28 PM UTC (45695fb)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1377,7 +1377,10 @@
     <tr>
       <th>Comments</th>
       <td>
-        If a descendant of a `table`, the first instance of a `caption` element will provide the `table` its accessible name.
+        <p class=note>
+          If a `caption` element is hidden from the accessibility tree, then it will not provide an accessible name
+          to its parent `table` element.
+        </p>
       </td>
     </tr>
   </tbody>
@@ -2296,7 +2299,7 @@
           <span class="type">Role:</span> Use WAI-ARIA mapping
         </div>
         <div class="relations">
-          <span class="type">Relations:</span> `IA2_RELATION_LABELLED_BY` with child <a href="#el-legend">`legend`</a> element
+          <span class="type">Relations:</span> `IA2_RELATION_LABELLED_BY` with first instance of child <a href="#el-legend">`legend`</a> element
         </div>
       </td>
     </tr>
@@ -2315,7 +2318,7 @@
           <span class="type">Role:</span> Use WAI-ARIA mapping
         </div>
         <div class="relations">
-          <span class="type">Relations:</span> `ATK_RELATION_LABELLED_BY` with child <a href="#el-legend">`legend`</a> element
+          <span class="type">Relations:</span> `ATK_RELATION_LABELLED_BY` with first instance of child <a href="#el-legend">`legend`</a> element
         </div>
       </td>
     </tr>
@@ -2339,6 +2342,10 @@
     <tr>
       <th>Comments</th>
       <td>
+        <p class=note>
+          If a `legend` element is hidden from the accessibility tree, then it will not provide an accessible name
+          to its parent `fieldset` element.
+        </p>
       </td>
     </tr>
   </tbody>
@@ -4972,6 +4979,10 @@
     <tr>
       <th>Comments</th>
       <td>
+        <p class=note>
+          If a `label` element is hidden from the accessibility tree, then it will not provide an accessible name
+          to the labelable element is is associated with.
+        </p>
       </td>
     </tr>
   </tbody>
@@ -7366,25 +7377,34 @@
     <tr>
       <th><a href="https://msdn.microsoft.com/en-us/library/dd373608%28v=VS.85%29.aspx">MSAA</a> + <a href="http://accessibility.linuxfoundation.org/a11yspecs/ia2/docs/html/">IAccessible2</a></th>
       <td>
-        Use WAI-ARIA mapping
+        <div class="role">Use WAI-ARIA mapping</div>
+        <div class="relations">
+          <span class="type">Relations:</span> `IA2_RELATION_LABELLED_BY` with first instance of child <a href="#el-caption">`caption`</a> element
+        </div>
       </td>
     </tr>
     <tr>
       <th><a href="https://msdn.microsoft.com/en-us/library/ms726297%28v=VS.85%29.aspx">UIA</a></th>
       <td>
-        Use WAI-ARIA mapping
+        <div class="role">Use WAI-ARIA mapping</div>
       </td>
     </tr>
     <tr>
       <th><a href="https://gnome.pages.gitlab.gnome.org/atk/">ATK</a></th>
       <td>
-        Use WAI-ARIA mapping
+        <div class="role">Use WAI-ARIA mapping</div>
+        <div class="relations">
+          <span class="type">Relations:</span> `ATK_RELATION_LABELLED_BY` with first instance of child <a href="#el-caption">`caption`</a> element
+        </div>
       </td>
     </tr>
     <tr>
       <th><a href="https://developer.apple.com/reference/appkit/nsaccessibility">AX</a></th>
       <td>
-        Use WAI-ARIA mapping
+        <div class="role">Use WAI-ARIA mapping</div>
+        <div class="role-namefrom">
+          <strong>AXDescription:</strong> value from child <a href="#el-caption">`caption`</a> subtree
+        </div>
       </td>
     </tr>
     <!-- <th><a href="#accessible-name-and-description-computation">Naming Algorithm</a></th> -->


fstrr (approved):
I did wonder about keeping the existing `caption` comment, but removing it is in line with the lack of similar comments in the document.

index.html
```diff
@@ -1377,7 +1377,10 @@ <h4 id=el-caption>`caption`</h4>
     <tr>
       <th>Comments</th>
       <td>
-        If a descendant of a `table`, the first instance of a `caption` element will provide the `table` its accessible name.
+        <p class=note>
+          If a `caption` element is hidden from the accessibility tree, then it will not provide an accessible name
+          to its parent `table` element.
```
adampage, line 1382:
```suggestion
          to its parent `table` element unless the `table` <a href="https://www.w3.org/TR/accname-1.2/#:~:text=user%20agents%20MUST%20include%20all%20nodes%20in%20the%20subtree%20as%20part%20of%20the%20accessible%20name%20or%20accessible%20description%2C%20when%20the%20node%20referenced%20by%20aria%2Dlabelledby%20or%20aria%2Ddescribedby%20is%20hidden">references it explicitly using `aria-labelledby`.</a>
```

If you can get past the ugliness of that deep link URL 🤦🏻‍♂️ , would it be too wordy or scope-creepy to call out [the `aria-labelledby` loophole](https://www.w3.org/TR/accname-1.2/#:~:text=user%20agents%20MUST%20include%20all%20nodes%20in%20the%20subtree%20as%20part%20of%20the%20accessible%20name%20or%20accessible%20description%2C%20when%20the%20node%20referenced%20by%20aria%2Dlabelledby%20or%20aria%2Ddescribedby%20is%20hidden) right here in each note?

adampage (changes requested):
Heya @scottaohara, I’ve just pushed up a some new WPT tests in web-platform-tests/wpt#44965 to support this PR. / cc @cookiecrook.

I hope I avoided the overlap and ballooning you cautioned against. 🤞🏻 Please let me know if you think they’re sensible/sufficient?

index.html
jnurthen: seems a very fragile way of linking. If we think we need to link to that we should create an idref to link to.

adampage: Yeah, I’ve been a little deep-link-happy lately. 😅

[`https://www.w3.org/TR/accname-1.2/#comp_labelledby`](https://www.w3.org/TR/accname-1.2/#comp_labelledby) would get folks to the right neighborhood instead.

scottaohara: i purposefully didn't add this into the notes, because it's not like using aria-labelledby makes the HTML element naming associations start working again.  Rather, the HTML naming associations never actually get trigged, since the aria-labelledby step has higher priority in the naming calculation.  

To put it another way, implementors aren't unclear about this aria-labelledby aspect now... this seems like it'd be more useful for authors to know.  So is this the best place to call this out?  I guess I'd be ok with a single note in the element naming section.  but, will authors come to this spec to understand this, or is this information better surfaced elsewhere?

scottaohara:
to be clear, the stuff i was really hoping for discussion on with the review of this PR was things like should hidden labels/legends/captions continue to even be associated with their respective elements, if they are hidden?  And thus, returning no name.

The example i brought up on the call was

```
<label hidden for=f>foo</label>
<input id=f title=blargh>
```

where the above example applies to tables and fieldsets with captions and legends, respectively.  

there is inconsistency in how the above is treated.  where some browsers return an empty name for the input.  And I believe I said it was Firefox returns "blargh" as the name of the input.

based on what this spec already has written for how these elements are named ([input naming as an example](https://w3c.github.io/html-aam/#input-type-text-input-type-password-input-type-number-input-type-search-input-type-tel-input-type-email-input-type-url-and-textarea-element-accessible-name-computation)) 

>- Otherwise use the associated label element or elements [accessible name(s)](https://www.w3.org/TR/accname-1.2/#dfn-accessible-name) - if more than one label is associated; concatenate by DOM order, delimited by spaces.
>- If the [accessible name](https://www.w3.org/TR/accname-1.2/#dfn-accessible-name) is still empty, then: use the control's title attribute.

My understanding is that Firefox is doing the right thing, and chromium/webkit are incorrectly stopping with the returned empty string from the hidden label/caption/legend.  

But since 2 browsers aren't following what the spec says, I am doubting myself if there's somehow a lack of clarity here.

adampage:
I agree, the “_If empty, then..._” statements are unambiguous and Firefox is doing the right thing.

In contrast, 4.1.2 has a more wobbly “[_Otherwise, if the control **still has no accessible name** use title attribute_](https://w3c.github.io/html-aam/#input-type-text-input-type-password-input-type-number-input-type-search-input-type-tel-input-type-email-input-type-url-and-textarea-element-accessible-name-computation:~:text=Otherwise%2C%20if%20the%20control%20still%20has%20no%20accessible%20name%20use%20title%20attribute.)”. For this one, Chrome and Webkit could argue that the control _does_ have an accessible name, it just happens to be empty.

Looking down the list of elements, I also see `input[type="image"]` has “_Otherwise use title attribute if present and its value is not the empty string_”, where the “_otherwise_” is similarly ambiguous. When step 1’s “_If the control has an... aria-labelledby..._” condition has been satisfied (even if the returned accname is empty), then implementers could be forgiven for thinking they can skip the rest?

So it does seem like there’s an opportunity to normalize all of the elements to that “_if empty_” phrasing so that `title` is always reliably used as the last resort.

index.html
adampage: Yes, got it. 👍🏻  I think the accname spec already describes this behavior well enough for authors. The mapping clarifications in this PR combined with the “_If still empty_” conditions in html-aam’s 4.1 seem clear enough to me.

adampage:
The [`figure`’s name computation](https://pr-preview.s3.amazonaws.com/w3c/html-aam/533/e644368...45695fb.html#figure-element-accessible-name-computation) of:

> 1. If...
> 2. If still empty...
> 3. If still empty...
> 4. Otherwise

...reads crystal clear to me. If you think normalizing all the 4.1 elements to this pattern would be sensible, I’d be happy to take a stab at that as a separate PR, and also add some [more WPT tests](https://github.com/web-platform-tests/wpt/pull/44965) for the `title` business.

fstrr:
Agree that Firefox is doing the right thing in using the `title` attribute.

Adam's suggestion to Normalize the 4.1 content to match the `figure` element's steps is a good idea.

index.html
scottaohara: glad that made sense.  thanks adam.  i'll leave this unresolved for a bit in case anyone else wants to weight in.  but otherwise i'll plan to mark this convo as resolved and not pull in this suggested change.

scottaohara:
i say let's just get that input consistency into this PR.  

we should probably do another PR / tests for the following, though

the table/fieldset already seem like they're clear at first glance.  but from what i've seen one could have

```
<fieldset>
  <legend hidden>foo</legend> <!-- this returns no name, so fieldset has no name -->
  <legend>test</legend> <!-- this is rendered as if there is no previous legend.  but it doesnt' contribute to name -->
</fieldset>
```

where the initial legend is hidden and an 'invalid' second legend is rendered instead.  visually, this looks just fine, but the accName  calc stops at the hidden legend, and doesn't get to the actually rendered legend to get a name from that.  Is that actually what people expect should happen here?

adampage:
Yeah, I see. Since the browser goes ahead and permissively renders that invalid second `<legend>`, the accname mapping should have equivalence. 😐

> ...if the `fieldset` element has a child that is a `legend` element, then use the subtree of the first such element.

In the above step from [4.1.5](https://pr-preview.s3.amazonaws.com/w3c/html-aam/533/e644368...45695fb.html#fieldset-element-accessible-name-computation), “_the first such element_” is plainly specific but also suggests a tolerance for multiple `legend` elements [even while HTML prohibits it](https://html.spec.whatwg.org/multipage/form-elements.html#the-legend-element:~:text=As%20the%20first%20child%20of%20a%20fieldset%20element.).

An allowance for invalid multi-`legend`/`caption` scenarios could look something like:

> ...if the `fieldset` element has a child that is a `legend` element **_and is exposed to the accessibility tree_**, then use the subtree of the first such element.

I guess we’re basically just doing our best to imitate whatever error correction algorithm that browsers are doing for rendering invalid HTML. Which I suppose is — please forgive my naivete — not standardized, or at least not specified anywhere that we could reference or copy?

scottaohara:
closing as now being handled in aria PR 2214

PR #538, opened 2024-03-19
Specify AXAPI behaviour for `low`, `high`, and `optimum` meter attributes.

### LINKED ISSUE #536: HTML AAM: VoiceOver (AXAPI) mapping for optimum, low <meter> attributes
Hi there, I'm working on VoiceOver support in Firefox and noticed the spec does not specify platform mappings for "low" and "optimum" HTML attributes on `<meter>` elements. 

In Safari, VoiceOver reads "suboptimum value" for something like:
`data:text/html,<label for="fuel">Fuel level:</label><meter id="fuel" min="0" max="100" low="33" high="66" optimum="80" value="50">at 50/100</meter>`

 "optimal value" for:
`data:text/html,<label for="fuel">Fuel level:</label><meter id="fuel" min="0" max="100" low="33" high="66" optimum="80" value="81">at 81/100</meter>`

and  "critical value" for:
`data:text/html,<label for="fuel">Fuel level:</label><meter id="fuel" min="0" max="100" low="33" high="66" optimum="80" value="10">at 10/100</meter>`

This string is appended to `AXValueDescription`, such that the total property reads, for example: "at 81/100, optimal value"

Can this mapping be added to the spec if it is official recommended practise?

MReschenberg:
Closes #536

Specification based on Safari behaviour/attributes exposed.

***
<a href="https://pr-preview.s3.amazonaws.com/MReschenberg/html-aam/pull/538.html" title="Last updated on Mar 20, 2024, 12:45 PM UTC (fcaaf19)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/html-aam/538/e644368...MReschenberg:fcaaf19.html" title="Last updated on Mar 20, 2024, 12:45 PM UTC (fcaaf19)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -11709,7 +11709,7 @@
     <tr>
       <th><a href="https://developer.apple.com/reference/appkit/nsaccessibility">AX</a></th>
       <td>
-        <div class="general">Not mapped</div>
+        <div class="general">Expose "optimal value" via `AXValueDescription` if &lt;value&gt; is greater than or equal to &lt;high&gt;. Expose "suboptial value" if &lt;value&gt; is less than &lt;high&gt; and greater than or equal to &lt;low&gt;.</div>
       </td>
     </tr>
     <tr>
@@ -12739,7 +12739,7 @@
     <tr>
       <th><a href="https://developer.apple.com/reference/appkit/nsaccessibility">AX</a></th>
       <td>
-        <div class="general">Not mapped</div>
+        <div class="general">Expose "suboptial value" via `AXValueDescription` if &lt;value&gt; is greater than or equal to &lt;low&gt; and less than &lt;high&gt;. If &lt;value&gt; is less than &lt;low&gt;, expose "critical value".</div>
       </td>
     </tr>
     <tr>
@@ -13995,7 +13995,7 @@
     <tr>
       <th><a href="https://developer.apple.com/reference/appkit/nsaccessibility">AX</a></th>
       <td>
-        <div class="general">Not mapped</div>
+        <div class="general">Expose "optimal value" via `AXValueDescription` if &lt;value&gt; is greater than or equal to &lt;optimum&gt;.</div>
       </td>
     </tr>
     <tr>
--- a/index.html
+++ b/index.html
@@ -11709,7 +11709,7 @@
     <tr>
       <th><a href="https://developer.apple.com/reference/appkit/nsaccessibility">AX</a></th>
       <td>
-        <div class="general">Expose "optimal value" via `AXValueDescription` if &lt;value&gt; is greater than or equal to &lt;high&gt;. Expose "suboptial value" if &lt;value&gt; is less than &lt;high&gt; and greater than or equal to &lt;low&gt;.</div>
+        <div class="general">Expose "optimal value" via `AXValueDescription` if &lt;value&gt; is greater than or equal to &lt;high&gt;. Expose "suboptimal value" if &lt;value&gt; is less than &lt;high&gt; and greater than or equal to &lt;low&gt;.</div>
       </td>
     </tr>
     <tr>


index.html
```diff
@@ -11709,7 +11709,7 @@ <h4 id=att-high>`high`</h4>
     <tr>
       <th><a href="https://developer.apple.com/reference/appkit/nsaccessibility">AX</a></th>
       <td>
-        <div class="general">Not mapped</div>
+        <div class="general">Expose "optimal value" via `AXValueDescription` if &lt;value&gt; is greater than or equal to &lt;high&gt;. Expose "suboptial value" if &lt;value&gt; is less than &lt;high&gt; and greater than or equal to &lt;low&gt;.</div>
```
jnurthen:
```suggestion
        <div class="general">Expose "optimal value" via `AXValueDescription` if &lt;value&gt; is greater than or equal to &lt;high&gt;. Expose "suboptimal value" if &lt;value&gt; is less than &lt;high&gt; and greater than or equal to &lt;low&gt;.</div>
```

jnurthen (changes requested):
I don't think it is as simple as this. As I understand it `<meter>` can have 3 segments, 
* (1) below `low`, 
* (2) between `low` and `high` 
* (3) above `high`. 
If the `optimum` value fits into (1) then all values within this segment are optimal, values in (2) are suboptimal and values in (3) are critical 
If the optimal value is in (2) then values in (1) and (3) are both suboptimal
If it is (3) then values in (1) are critical, (2) suboptimal and (3) optimal.

These correspond to the visual coloring of green (optimal), orange (suboptimal) and red (critical). 

If no `optimum` value is specified then the optimal segment is segment (2)

Here is the webkit logic https://github.com/WebKit/WebKit/blob/c330ea99229f6bb478dde74497ee652e44a3d7d5/Source/WebCore/html/HTMLMeterElement.cpp#L156

index.html
```diff
@@ -11709,7 +11709,7 @@ <h4 id=att-high>`high`</h4>
     <tr>
       <th><a href="https://developer.apple.com/reference/appkit/nsaccessibility">AX</a></th>
       <td>
-        <div class="general">Not mapped</div>
+        <div class="general">Expose "optimal value" via `AXValueDescription` if &lt;value&gt; is greater than or equal to &lt;high&gt;. Expose "suboptimal value" if &lt;value&gt; is less than &lt;high&gt; and greater than or equal to &lt;low&gt;.</div>
```
rahimabdi, line 11712: Relevant WebKit code: https://github.com/WebKit/WebKit/blob/c330ea99229f6bb478dde74497ee652e44a3d7d5/Source/WebCore/html/HTMLMeterElement.cpp#L172

This looks correct for `high`. We'll also need to add the case for when the `value` is less than "low" (for which this is exposed as "critical value").

index.html
```diff
@@ -12739,7 +12739,7 @@ <h4 id=att-low>`low`</h4>
     <tr>
       <th><a href="https://developer.apple.com/reference/appkit/nsaccessibility">AX</a></th>
       <td>
-        <div class="general">Not mapped</div>
+        <div class="general">Expose "suboptial value" via `AXValueDescription` if &lt;value&gt; is greater than or equal to &lt;low&gt; and less than &lt;high&gt;. If &lt;value&gt; is less than &lt;low&gt;, expose "critical value".</div>
```
rahimabdi, line 12742: Relevant WebKit code: https://github.com/WebKit/WebKit/blob/c330ea99229f6bb478dde74497ee652e44a3d7d5/Source/WebCore/html/HTMLMeterElement.cpp#L163

For `low`:

- If `value` is less than or equal to `low`, then `value` is "optimal value"
- if `value` is less than or equal to `high` (but greater than `low`), then `value` is "suboptimal value"
- else, `value` is "critical value"

Also, looks like there is a typo for the word "suboptimal" here.

index.html
```diff
@@ -13995,7 +13995,7 @@ <h4 id=att-optimum>`optimum`</h4>
     <tr>
       <th><a href="https://developer.apple.com/reference/appkit/nsaccessibility">AX</a></th>
       <td>
-        <div class="general">Not mapped</div>
+        <div class="general">Expose "optimal value" via `AXValueDescription` if &lt;value&gt; is greater than or equal to &lt;optimum&gt;.</div>
```
rahimabdi, line 13998: Relevant WebKit code: https://github.com/WebKit/WebKit/blob/c330ea99229f6bb478dde74497ee652e44a3d7d5/Source/WebCore/html/HTMLMeterElement.cpp#L184

For `optimum`:

- If `value` is greater than or equal to `low` AND `value` is less than or equal to `high` then `value` is  "optimal value"
- Else, `value` is "suboptimal value"

Note that in this case, there is no "critical value".

rahimabdi (commented):
Hi @MReschenberg, I've added comments and recommendations that largely align with @jnurthen's assessment of `optimum` treatment.

cookiecrook (changes requested):
This issue should have an active test case showing this in a more common use case, like an audio meter bouncing quickly and showing critical peak line clipping. 

I’m not sure if this will need new API, or if there is existing API, but I’m reasonably sure `AXValueDescription` is not the ideal way to vend this to AT like VO.

Update: dynamic meter test case added to a [new WebKit Accessibility bug](https://webkit.org/b/273589).

MReschenberg:
> I’m not sure if this will need new API, or if there is existing API, but I’m reasonably sure `AXValueDescription` is not the ideal way to vend this to AT like VO.

Do you have an alternative attribute in mind? 
It sounds like there's not enough info at this point to make this spec decision, so I'm leaning towards updating firefox's VO implementation to match safari temporarily and leaving this open for discussion.

cookiecrook:
Role/Subrole is already `AXLevelIndicator/AXMeter` but other attrs TBD based on the outcome of the linked issue.

rahimabdi:

- https://github.com/w3c/aria/pull/2339

PR #609, opened 2026-07-29
set up redirect

daniel-montalvo:


tidoust:
@daniel-montalvo, this redirect works but loses fragments along the way, creating broken links in specs that reference sections in the html-aam spec.

I filed https://github.com/WICG/aom/pull/213 to update the URL in AOM. There's also a broken link in HTML Ruby Markup Extensions.

An alternative would be to also add a short `onload` script to preserve the fragment as done when we do redirects at the W3C level (for example, see the redirect for [a11y-discov-vocab](https://github.com/w3c/w3c.github.io/blob/main/a11y-discov-vocab/index.html#L7C1-L7C97). Something like:

```html
<body onload='location.href = "https://w3c-cg.github.io/a11y-discov-vocab/" + location.hash;'>
```

daniel-montalvo:
> @daniel-montalvo, this redirect works but loses fragments along the way, creating broken links in specs that reference sections in the html-aam spec.
> 
> I filed [WICG/aom#213](https://github.com/WICG/aom/pull/213) to update the URL in AOM. There's also a broken link in HTML Ruby Markup Extensions.
> 
> An alternative would be to also add a short `onload` script to preserve the fragment as done when we do redirects at the W3C level (for example, see the redirect for [a11y-discov-vocab](https://github.com/w3c/w3c.github.io/blob/main/a11y-discov-vocab/index.html#L7C1-L7C97). Something like:
> 
> ```
> <body onload='location.href = "https://w3c-cg.github.io/a11y-discov-vocab/" + location.hash;'>
> ```

Will do.

